PR #10, opened 2021-09-09
Add gas in args

LucienSong:
Add gas in args to mute the console warning


LucienSong:
I  had signed the CLA already but the status is still pending.

PR #31, opened 2021-11-11
webgl on network change update

leon-do:
- rm 2021 template
- Added method to get network id


KBryan:
Looks good. Now that we have removed 2020. We may want to rename our build 202x or something in the future as 2020 templates are good anything higher.

PR #42, opened 2021-12-06
Graph prefabs

RyRy79261:
Pending export

### DIFF
--- a/Web3Unity/Scripts/Library/EVM.cs
+++ b/Web3Unity/Scripts/Library/EVM.cs
@@ -8,8 +8,6 @@ using Newtonsoft.Json;
 public class EVM
 {
   public class Response<T> { public T response; }
-  // public class StringResponse { public string response; }
-  // public class IntResponse { public int response; }
 
   private readonly static string host = "https://api.gaming.chainsafe.io/evm";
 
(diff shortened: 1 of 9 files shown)


Web3Unity/Scripts/Library/EVM.cs
```diff
@@ -123,8 +122,35 @@ public static async Task<string> CreateContractData(string _abi, string _method,
...
+  {
+    WWWForm form = new WWWForm();
+    form.AddField("chain", _chain);
+    form.AddField("network", _network);
+    form.AddField("account", _account);
+    form.AddField("contract", _contract);
+    string url = host + "/all1155";
+    UnityWebRequest webRequest = UnityWebRequest.Post(url, form);
+    await webRequest.SendWebRequest();
+    Response<string> data = JsonUtility.FromJson<Response<string>>(System.Text.Encoding.UTF8.GetString(webRequest.downloadHandler.data));
```
leon-do, line 153: this one is a bit tricky. The response is not a string but something like like:

```
[
        {
            "contract": "0x7dca125b1e805dc88814aed7ccc810f677d3e1db",
            "tokenId": "24",
            "uri": "https://opensea-creatures-api.herokuapp.com/api/creature/24",
            "balance": "1"
        }
]
```

PR #84, opened 2022-01-27
add remember me toggle and rename files to web3wallet

leon-do:


leon-do:
Please test Web3Unity > Scenes > WalletLogin

KBryan:
The current implementation doesn't work as expected. I'm moved to the sign message copy but doesn't move to the next scene.

PR #182, opened 2022-07-06
ftm and cro chains added

sneakzttv:
Ftm and Cro chains added to Web3PrivateKey.cs for Kwame to review.


sneakzttv:
it's trying to pull the sample scene in with it, just want to check that this is the correct way of pushing a build / things i need to fix etc. Do i just delete the sample scene usually?

sneakzttv:
closing as i want to try again without sample scene

PR #183, opened 2022-07-06
web3gl.lib update

sneakzttv:
console errors fixed in web3gl.jslib
extra chains added to web3privatekey.cs
mobile sign added to web3gl.jslib/web3gl.cs


sneakzttv:
closed, creating a new one with metadata removed.

PR #193, opened 2022-07-19
Lazy minter prefab

sneakzttv:
Branch created, prefabs created and tested. Returning both calls on button press. Bugs out sometimes though from the server response.


KBryan (changes requested):
Looks good. We can remove the minter.cs file as you have wired up that functionality in the prefabs now.

sneakzttv:
ok will do, i've noticed when pulling the branch the scripts don't stick to the prefab even though i've assigned them. I'll double check this too.

PR #196, opened 2022-07-22
Lazyminterprefab

sneakzttv:
- Webwallet Voucher Minter Prefabs added
- Webgl Voucher Minter Prefabs added
- Created sub folders for both of these in the minter folder as there were lots of prefabs
- Both systems are saving the vouchers to playerprefs in case the user closes the game
- Once validated that the signer is correct against playerprefs, mint is used the voucher is nullified
- Added comments should the user wish to alter the code to fit their system


sneakzttv:
metadata files aren't an issue, annoying to go through and delete each one anytime i change anything xD

KBryan (approved):
Looks good

PR #231, opened 2022-09-22
Updated URI class method

KBryan:
Modified class method to handle tokenId hex returned from minted assets in marketplace.


sneakzttv:
perfect, merging

PR #232, opened 2022-09-22
Updated ImportNFTTextureExample to work with ChainSafe MarketPlace

KBryan:


sneakzttv:
chain changed to goerli and fallback added to ipfs, good job, works

PR #264, opened 2022-11-16
1.6.0 Mint alterations

sneakzttv:
Mint alterations

- Minting without a voucher removed from scripts.
- Chicken demo 1155 voucher example added for users to understand the process as well as tuple/array formats.


sneakzttv:
Unsure if we should leave PK and IPFS Minting

sneakzttv:
hang on that wants to go to main, no

sneakzttv:
commits have been added to 1.6.0 <3

PR #324, opened 2023-02-13
Added Task.FromResult to RPC call

KBryan:


sneakzttv (approved):
noice no more wanrings

PR #335, opened 2023-02-18
Ethers core

KBryan:


sneakzttv (approved):
core dll woo

Scenes/SampleScene.unity
```diff
@@ -0,0 +1,303 @@
+%YAML 1.1
```
FSM1, line 1: Any reason this is being added to the repo?

Resources/BillingMode.json
```diff
@@ -0,0 +1 @@
+{"androidStore":"GooglePlay"}
```
FSM1, line 1: Same here

FSM1, line 1: Same here

FSM1, line 1: Any reason this is being added to the repo?

FSM1 (changes requested):
A couple of files made it in to the repo that I think don't necessarily have to be there.

sneakzttv (approved):

PR #342, opened 2023-02-24
analytics fix

sneakzttv:
analytics fix


KBryan (commented):
I'm looking at this, and it seems like you are missing getters and setters for the values. 

public string projectId
{
	get { return ProjectID; }
	set { ProjectID = value; }
}

Or something to that effect, taking into account the syntax you provided.

sneakzttv:
fixed @KBryan @juans-chainsafe

KBryan (approved):

juans-chainsafe (approved):
Working on WebGL build and transition scenes is working as well!

PR #363, opened 2023-03-28
Autogeneration

sneakzttv:
auto generation was leaving out a file, included it explicitly and it works.


juans-chainsafe (approved):

KBryan (commented):
Shouldn't these be autogenerated? They are now hard coded.

KBryan (approved):

PR #370, opened 2023-04-07
Update index.html

sneakzttv:
figured out that the scripts loading too early was the issue with webgls boot, pushed them both to the bottom of the body in the html file and things are loading fine now in testing.


juans-chainsafe (approved):
Looks good to me!

KBryan (approved):

PR #371, opened 2023-04-10
Webgl 2nd fix

sneakzttv:
Upon further testing over the weekend i realized that async doesn't work in the body it has to be in the head as some users were getting errors on older browsers. I shifted it to the head but was still getting the error so i moved all the scripts to the bottom of the body without async. Essentially it's loading the html first then loading each script in order after that as opposed to how we had it before which was at the top. The commits are me testing variations, sorry for the mess.

### DIFF
--- a/Assets/WebGLTemplates/Web3GL-2020x/index.html
+++ b/Assets/WebGLTemplates/Web3GL-2020x/index.html
@@ -24,13 +24,6 @@
       </div>
     </div>
     <script>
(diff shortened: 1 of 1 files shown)


juans-chainsafe (approved):
I manage to generate a webgl build with Login without issues

KBryan:
@juans-chainsafe are you also testing for builds deployed on self-hosting?

juans-chainsafe:
> @juans-chainsafe are you also testing for builds deployed on self-hosting?

no, not sure how I can test that using the code on this branch

KBryan (approved):

Issue #375, opened 2023-04-11
Removing cronos template


sneakzttv:
Removing the cronos template from the SDK as v2 encapsulates all of it. It's also bulking out the network.js file.

PR #376, opened 2023-04-11
Cronos removal & wallet connect fix

sneakzttv:
Cronos SDK removed as it's going to become obsolete when the api is shutdown. Wallet connect bug that was causing an account to stay logged in fixed by adding a disconnect call at wallet connects initialization step.


sneakzttv:
closes #375

RyRy79261 (approved):

PR #381, opened 2023-04-14
Webgl boot error

### LINKED ISSUE #374: Webgl boot error
Webgl is having issues loading the scripts sometimes and just errors ou t. A refresh usually gets over this and I've been through a few iterations with defer and async that load slightly better, but they still error on some older browsers. I'd like to make this cleaner. You can see this here if you spam refresh https://chainsafe.github.io/ChainSafe-Demo-WebGL/

sneakzttv:
seems slightly better, closes #374 until it pops up again


sneakzttv:
if all good we'll create a new release @juans-chainsafe and get the other fixes out also. At least we only have 1 ting to fix it pops up later.

KBryan:
It is best to assemble all the edge cases for the WebTamplate update and do an end-to-end test against it.

KBryan:
Okay, I first noticed that it doesn't autodetect the network. I can log into the chicken demo on mainnet and interact with the game. The old version of the WebTemplate would switch you to the correct network, which used to be goerli. Did this requirement change in the new release?

KBryan:
I don't see this error and not able to reproduce. Will close out this issue and revisit if it comes back up.

KBryan:
It looks like a user has brought up this issue also. @nftpixels You mentioned that you are seeing this also. Going to reopen this issue.
https://discord.com/channels/593655374469660673/845686120292679761/1099107404060307456

Issue #423, opened 2023-05-16
GetArray prefab not working on WebGL


juans-chainsafe:
**Describe the bug**
GetArray is not working on WebGL builds, but is working on Web3Wallet, looks like the error is on the JsonRCPProvider that you need to manually change on the WebGLGetArray.cs file

**To Reproduce**
1. Add the LoginWallet scene
2. Add the prefab 'WebGLReturnArrays'
3. Generate a WebGL build
4. Do the login
5. Click on 'Set Array Data' button and wait for transaction to finish
6. Click on 'Get Array Data' button
7. See error

**Expected behavior**
First address stored is displayed on the screen

**Console logs**
2 different errors here:
If I try with var provider = RPC.GetInstance.Provider(); I get: exception thrown: `RuntimeError: null function or function signature mismatch,RuntimeError: null function or function signature mismatch`
If I try with var provider = `new JsonRpcProvider("https://goerli.infura.io/v3/904006115c764661965dc0909e5ed473");` (my provider) I get: `(methodAccessException: Attempt to access method ‘Web3Unity.Scripts.Library.Ethers.RPC.IRpcEnvironment.LogError’ on type ‘’ failed.)
`

**Desktop (please complete the following information):**
 - Unity Platform: [WebGL ]
 - Unity version [2021.3.18] 
 - OS [Mac]

juans-chainsafe:
> So, this works if you use `RPC.GetInstance.Provider();` or setting your RPC provider manually. This would make sense in WebGL as PlayerPrefs do always work as intended in WebGL builds.

Actually it doesn't work for me if I set it up manually, the error that I get is putting my rcp provider manually

KBryan:
Okay, thanks for clarifying this.

sneakzttv:
prefab is missing an event system, updating the branch to include this as you can't click the buttons without it. I'm seeing his error also.

sneakzttv:
found the fix :)

PR #426, opened 2023-05-16
webgl-prefab-fix

sneakzttv:
Using the RPC getter function fixes this after an event system was added. Working on my end.


sneakzttv:
Closes #423

juans-chainsafe (approved):
Thanks for fixing this! GetArray is working as expected nooow.

Also, I added the abi and method in the Web3WalletTransfer1155.example so we don't need to add it manually every time.

KBryan (approved):

PR #427, opened 2023-05-16
WebGL Template MetaMask

KBryan:
- WebGL Template with only MetaMask as dependency


sneakzttv (approved):

sneakzttv:
i think this got missed in the migration, good spot!

Arshia001 (approved):

PR #429, opened 2023-05-19
Feature/add token contract

KBryan:
Feature to give Unity developer the ability to add a custom token to EOA if not already present.


sneakzttv:
oh hey this is cool!

PR #432, opened 2023-05-20
feat/webgl fix & nft api

sneakzttv:
This fixes the webgl initialization error as well as integrating the infura api fix from @nftpixels for the allnft calls.
@juans-chainsafe & @nftpixels please review when you get time on monday <3


nftpixels (approved):
Great to see a replacement for the graph coming in! :)

sneakzttv:
Waiting for @kalambet to attempt a solution as this isn't the way to fix a critical bug even though it's fully functional & tested.

Arshia001 (changes requested):
Minified file should not be changed directly. Do we have the un-minified sources?

RyRy79261:
@Arshia001 The one being used is from this [branch of the template repo](https://github.com/ChainSafe/web3.unity-webgl-template/tree/feat/parcel-refactor)

Arshia001:
@RyRy79261 so if we figure out where exactly the error is, we can just make the same change in that repo, right?

PR #450, opened 2023-06-01
Update IEvmSigner

KBryan:
Extend IEvmSigner to have a method to derive an address from a private key using a string as a param.


sneakzttv (approved):

sneakzttv:
needed for w3a

KBryan:
@Arshia001 && @oleksandrchainsafe can you take a look at this? Currently blocked.

RyRy79261 (approved):

Arshia001:
I don't know how I never noticed this PR, but this functionality doesn't belong in the Signer. Signers don't always have access to their private keys, and even if they did, since the pk is an input parameter, it doesn't have to live on the signer at all; this can be a static method somewhere (unless I'm understanding the whole thing incorrectly). @KBryan please revert this PR and move this function to a static utility class.

PR #453, opened 2023-06-02
Feature/project configurations

oleksandrchainsafe:


src/UnityPackage/Assets/Plugins/Web3/Scripts/ProjectConfigScriptableObject.cs
```diff
@@ -35,9 +37,11 @@ public string Network
...
     }
 
-    public string RPC
+    public string Rpc
     {
         get => rpc;
         set => rpc = value;
     }
+
+    public static ProjectConfigScriptableObject LoadDefault() => Resources.Load<ProjectConfigScriptableObject>(DefaultAssetPath);
```
Arshia001, line 46: I have an entire static class dedicated to loading and saving the project config in the UPM PR. Let's put this PR on top of that and use that class.

src/UnityPackage/Assets/Web3Unity/Scripts/Scenes/WebLogin.cs
```diff
@@ -22,12 +22,12 @@ public class WebLogin : MonoBehaviour
     void Start()
     {
         // loads the data saved from the editor config
-        projectConfigSO = (ProjectConfigScriptableObject)Resources.Load("ProjectConfigData", typeof(ScriptableObject));
-        PlayerPrefs.SetString("ProjectID", projectConfigSO.ProjectID);
-        PlayerPrefs.SetString("ChainID", projectConfigSO.ChainID);
+        projectConfigSO = ProjectConfigScriptableObject.LoadDefault();
+        PlayerPrefs.SetString("ProjectID", projectConfigSO.ProjectId);
```
Arshia001, line 26: Do we even need the player prefs any more? I suppose we have to keep them so as to stay backwards-compatible, but we don't need them in the core of the SDK, only the compat layer.

Arshia001 (changes requested):

src/UnityPackage/Assets/Web3Unity/Scripts/Scenes/WebLogin.cs
sneakzttv: we shouldn't need them there if they're being loaded in elsewhere and said new class is being referenced in the appropriate calls.

sneakzttv:
linter has failed also

src/UnityPackage/Assets/Web3Unity/Scripts/Scenes/WebLogin.cs
oleksandrchainsafe: I didn't change anything here as it's out of the scope if this task

sneakzttv (dismissed):

src/UnitySampleProject/Assets/Scripts/Scenes/WebLogin.cs
```diff
@@ -22,9 +22,9 @@ public class WebLogin : MonoBehaviour
     void Start()
     {
         // loads the data saved from the editor config
-        projectConfigSO = (ProjectConfigScriptableObject)Resources.Load("ProjectConfigData", typeof(ScriptableObject));
-        PlayerPrefs.SetString("ProjectID", projectConfigSO.ProjectID);
-        PlayerPrefs.SetString("ChainID", projectConfigSO.ChainID);
+        projectConfigSO = ProjectConfigScriptableObject.LoadDefault();
```
Arshia001, line 25: This has to be `ProjectConfigUtilities.Load()`. I think you made the same mistake as me: not having WebGL as the active platform when working on the codebase.

PR #456, opened 2023-06-06
Feat/network js fix

sneakzttv:
Automatically writes the network.js files on ProjectID save so the users no longer need to!

### DIFF
--- a/src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
+++ b/src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
@@ -149,10 +149,10 @@ public class ChainSafeServerSettings : EditorWindow
         writer1.WriteLine("//You can see a list of compatible EVM chains at https://chainlist.org/");
         writer1.WriteLine("window.networks = [");
         writer1.WriteLine("     {");
-        writer1.WriteLine("             id: " + PlayerPrefs.GetString("ChainID") + ",");
-        writer1.WriteLine("             label: " + '"' + PlayerPrefs.GetString("Chain") + " " + PlayerPrefs.GetString("Network") + '"' + ",");
-        writer1.WriteLine("             symbol: " + '"' + PlayerPrefs.GetString("Symbol") + '"' + ",");
-        writer1.WriteLine("             rpcUrl: " + "'" + PlayerPrefs.GetString("RPC") + "'" + ",");
+        writer1.WriteLine("            id: " + PlayerPrefs.GetString("ChainID") + ",");
+        writer1.WriteLine("            label: " + '"' + PlayerPrefs.GetString("Chain") + " " + PlayerPrefs.GetString("Network") + '"' + ",");
+        writer1.WriteLine("            symbol: " + '"' + PlayerPrefs.GetString("Symbol") + '"' + ",");
+        writer1.WriteLine("            rpcUrl: " + "'" + PlayerPrefs.GetString("RPC") + "'" + ",");
         writer1.WriteLine("     }");
         writer1.WriteLine("]");
         writer1.Close();
--- a/src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
+++ b/src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using UnityEngine;
 using SDKConfiguration;
 using System;
+using System.IO;
 
 public class ChainSafeServerSettings : EditorWindow
 {
@@ -10,6 +11,7 @@ public class ChainSafeServerSettings : EditorWindow
     public string ChainID = "Please Enter Your Chain ID";
     public string Chain = "Please Enter Your Chain i.e Ethereum, Binance, Cronos";
     public string Network = "Please Enter Your Network i.e Mainnet, Testnet";
+    public string Token = "Please Enter Your Chain's Native Token i.e Eth, Cro";
     public string RPC = "Please Enter Your RPC";
     public User saveObject;
     Texture2D m_Logo = null;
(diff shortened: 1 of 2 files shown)


src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
```diff
@@ -125,4 +137,30 @@ void OnGUI()
...
+        // declares paths to write our javascript files to
+        string path1 = "Assets/WebGLTemplates/Web3GL-2020x/network.js";
+        string path2 = "Assets/WebGLTemplates/Web3GL-MetaMask/network.js";
+
+        // writes data to the webgl default network file
+        StreamWriter writer1 = new StreamWriter(path1, false);
+        writer1.WriteLine("//You can see a list of compatible EVM chains at https://chainlist.org/");
+        writer1.WriteLine("window.networks = [");
+        writer1.WriteLine("  {");
+        writer1.WriteLine("id: " + PlayerPrefs.GetString("ChainID") + ",");
```
Arshia001, line 152: These need to be indented with 4 spaces.

sneakzttv: tab only gave me 2 for some reason, odd

sneakzttv: moment

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
```diff
@@ -125,4 +137,30 @@ void OnGUI()
         }
         GUILayout.Label("Reminder: Your ProjectID Must Be Valid To Save & Build With Our SDK. You Can Register For One On Our Website At Dashboard.Gaming.Chainsafe.io", EditorStyles.label);
     }
+
+    public void WriteNetworkFile()
+    {
+        // declares paths to write our javascript files to
+        string path1 = "Assets/WebGLTemplates/Web3GL-2020x/network.js";
```
Arshia001, line 144: The templates shouldn't be hard-coded, the code should list the folders inside `WebGLTemplates` and create the file in all folders that are actually there.

Arshia001 (changes requested):

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
sneakzttv: how do you mean?

sneakzttv: or this 

```
string root = "Assets/WebGLTemplates/";
var info = new DirectoryInfo(root);
var fileInfo = info.GetFiles();
```

sneakzttv: also if you noticed, both network files are different, so i'm not sure if a for x for each file work

sneakzttv:
@juan Currently you'll see a newtonsoft error when you pull main, i have no idea why but this fixes it. @Arshia001 you may want to look into this too i'm not sure who or what caused it, this error was present when i started.

1: Create new references
image

2: remove unused references
image

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
sneakzttv: fixed

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
```diff
@@ -43,6 +45,12 @@ void Awake()
             PlayerPrefs.Save();
         }
 
+        if (Token == ("Please Enter Your Chain's Native Token i.e Eth, Cro") && (PlayerPrefs.GetString("Token") != ""))
```
RyRy79261, line 48: I think the terminology here is symbol, consider rephrasing to: 
```suggestion
        if (Symbol == ("Please Enter Your Chain's Native Symbol i.e Eth, Cro") && (PlayerPrefs.GetString("Symbol") != ""))
```

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
RyRy79261: @sneakzttv @Arshia001 might mean with regards to `Web3GL-2020x`, there might be additional or different version templates in there, meaning this should be potentially created in all templates that appear to require the file, not just the one expected.

If I'm reading that correctly

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
RyRy79261: Not sure if I missed the fix, but I can suggest them here, would need to be done to label, token & rpc url probably 
```suggestion
        writer1.WriteLine("    id: " + PlayerPrefs.GetString("ChainID") + ",");
```

RyRy79261 (changes requested):
A clean implementation besides a semantic question and indentation. But in terms of maintainability, why not create a `Class` for the server setting parameters, then serialize that directly into a file?

That way we're sticking to writing C# as much as possible and not json inside of string quotes

[Example 1](https://stackoverflow.com/questions/41427643/c-sharp-serialize-and-deserialize-json-to-txt-file)

[Example 2](https://stackoverflow.com/questions/41427643/c-sharp-serialize-and-deserialize-json-to-txt-file)

src/UnityPackage/Assets/Editor/ServerSettings/ChainsafeServerSettings.cs
sneakzttv: like that interms of indentation and symbol?

RyRy79261: @sneakzttv I mean the term for the chain's local token is usually referred to as "Symbol"

For example, the fields in Metamask
image

sneakzttv: yah got it, updated the rest, left the object as is

juans-chainsafe (approved):
Tried a few prefabs with WebGL and Webwallet, the player settings where save dinto the network.js file correctly with the symbol!

KBryan (approved):

PR #457, opened 2023-06-07
Revert "Update IEvmSigner (#450)"

Arshia001:
This reverts commit 82bf35f949423733f9f6cb38a86cb209798c27f4.


sneakzttv:
why are we reverting this one? it's needed for the way web3auth gets the wallet address.

sneakzttv (approved):

sneakzttv:
After deliberations with Kwame, this can safely be removed as we're now on the same page.

KBryan (approved):

PR #458, opened 2023-06-07
Feature/lifecycle participation

Arshia001:


Arshia001:
Note: this PR can only be merged after #447. All commits since fde398f must be rebased before merging.

sneakzttv:
@juans-chainsafe round 2 beep boop

Arshia001:
This code does not work as-is, it will cause a deadlock when running the game.

sneakzttv (approved):
mostly async and lifecycle alterations, looks good to me.

src/UnitySampleProject/Assets/Scripts/Prefabs/Wallet/Web3WalletGetBlockNumber.cs
KBryan, line 1: Curious as to why functionality is being removed from a prefab. Are we no longer shipping prefabs and if so. You should just remove the script altogether.

src/UnitySampleProject/Assets/Scripts/Prefabs/Wallet/Web3WalletGetGasLimit.cs
KBryan, line 1: Curious as to why functionality is being removed from a prefab. Are we no longer shipping prefabs and if so. You should just remove the script altogether.

KBryan (changes requested):
There are multiple instances where I see functionality being commented out instead of implemented for its purposes. If this code is no longer needed, you should delete it along with its associated prefab and corresponding metadata.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Web3.cs
```diff
@@ -51,29 +52,29 @@ private static T AssertComponentAccessible<T>(T value, string propertyName)
...
-        public async ValueTask Initialize()
+        public async ValueTask InitializeAsync()
         {
             if (initialized)
             {
                 throw new Web3Exception("Web3 was already initialized.");
             }
 
-            if (provider != null)
+            foreach (var lifecycleParticipant in serviceProvider.GetServices<ILifecycleParticipant>())
```
oleksandrchainsafe, line 68: This can be a separate service, just to keep web3 cleaner. This way web3 could be treated as a facade delegating all of the work to its components.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Web3.cs
```diff
@@ -82,7 +83,10 @@ public void Terminate()
 
             if (initialized)
             {
-                // todo terminate other components
+                foreach (var lifecycleParticipant in serviceProvider.GetServices<ILifecycleParticipant>())
```
oleksandrchainsafe, line 86: same with this

oleksandrchainsafe (changes requested):

src/UnitySampleProject/Assets/Scripts/Prefabs/ERC20/ERC20CustomTokenBalance.cs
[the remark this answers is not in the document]
Arshia001: So, this is not a complete version of the code, just a WIP. The prefabs are coming back, so we're keeping the code there to migrate it to the new API later (there's a ticket for that). For now, we have to comment them out since the supporting code is simply not there.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Web3.cs
Arshia001: Lifecycle participation is already its own service. If you mean the part where it's calling the implementations, it really can't be a separate service since `Web3` would have to call that other service anyway, there's simply no other place to be calling the code from. It'd be a layer of indirection that doesn't really do anything.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Web3.cs
Arshia001: Gonna resolve this one, let's continue the discussion on the other one

KBryan (approved):

src/ChainSafe.GamingSDK.EVM/Web3/Core/Web3.cs
Arshia001: You know, I kind of feel we're over-engineering here. Too much code in one class is bad, but too many classes with few lines of actual logic is also bad for readability. Let's keep this, and when it does make sense to have a separate service, we'll add it in.

oleksandrchainsafe (approved):

PR #460, opened 2023-06-08
Use Paket as package manager

Arshia001:
This PR replaces our direct use of Nuget with Paket, which is an alternative package manager for dotnet that works the same way that cargo or npm do: You specify a list of dependencies with version constraints, then the package manager figures out the entire dependency tree and keeps it locked in and keeps the versions similar between projects.

Note that after this change is merged in, nuget references should no longer be added to projects directly. Instead, they should be added to the `paket.dependencies` file that lives in the root of the repo and for each project that needs those dependencies, they should also be added to the `paket.references` file of that project.

After making changes to either file, run `dotnet paket install` and the dependency tree will be recalculated and the correct packages will be restored and be ready for use.

After cloning this change for the first time, the `setup.bat` script in the root of the repo should be run once.


sneakzttv:
1. git submodule init
2. git submodule update
3. run setup file
4. run package batch file
5. open web3.unity\src\UnitySampleProject from unity hub

receiving these errors
image

juans-chainsafe (changes requested):
I deleted my web3.unity locally and cloned the repository from a fresh start. Then I did:

1. dotnet clean ✅ 
2. sh setup.sh ✅ 
3. sh src/ChainSafe.GamingSdk.EVM.Unity/publish-to-unity-package.sh ❌ 

Error on step 3:
`/usr/local/share/dotnet/sdk/7.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(266,5): error NETSDK1005: Assets file '/Users/juanmanuelspoleti/Desktop/workspace/web3.unity/src/ChainSafe.GamingSDK.EVM.Tests/obj/project.assets.json' doesn't have a target for 'netstandard2.1'. Ensure that restore has run and that you have included 'netstandard2.1' in the TargetFrameworks for your project. [/Users/juanmanuelspoleti/Desktop/workspace/web3.unity/src/ChainSafe.GamingSDK.EVM.Tests/ChainSafe.GamingSDK.EVM.Tests.csproj]`

sneakzttv:
oh i thought Arshia said we did need steps 1 and 2 in standup, I must've misheard that.

Arshia001:
@juans-chainsafe @sneakzttv I have reproduced the issue. It wasn't showing up on my system because the exitance of previous versions of the code affects whether it happens. I'm guessing it has something to do with the contents of one of the `obj` folders. I'm going to stop working on this for a while, since it can get pretty difficult to debug and doesn't offer huge gains.

Arshia001:
Closing this PR and moving the issue back to backlog since this is not going to be worked on in the near future.

PR #464, opened 2023-06-12
Feature/analytics client

oleksandrchainsafe:
Analytics Client that sends events to the API server and is integrated into the new architecture.


src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/ApiAnalyticsClient.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.Generic;
+using ChainSafe.GamingWeb3.Environment;
+
+namespace ChainSafe.GamingWeb3.Analytics
+{
+    public class ApiAnalyticsClient : IAnalyticsClient
+    {
+        private const string LoggingUrl = "https://game-api-stg.chainsafe.io/logging/logEvent";
```
Arshia001, line 8: Why are we using the staging API here? AFAIK, the API is also available on production, so let's use that.

kalambet, line 8: Is it possible not to hardcode any URLs but read them from settings, configs, or environment?

src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/ApiAnalyticsClient.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        private readonly IProjectConfig projectConfig;
+        private readonly IChainConfig chainConfig;
+        private readonly IHttpClient httpClient;
+        private readonly ILogWriter logWriter;
```
Arshia001, line 13: I don't think this is actually being used?

src/ChainSafe.GamingSdk.EVM.Unity/UnityEnvironmentExtensions.cs
```diff
@@ -17,18 +18,18 @@ public static IWeb3ServiceCollection UseUnityEnvironment(this IWeb3ServiceCollec
 
         public static IWeb3ServiceCollection UseUnityEnvironment(this IWeb3ServiceCollection services)
         {
+            services.UseApiAnalytics();
```
Arshia001, line 21: We should probably document that we enable analytics by default, and that people can opt out by calling `DisableAnalytics`.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Evm/Dto/Chains.cs
```diff
@@ -49,7 +49,7 @@ private static void CaptureError(string error, Web3Environment environment)
                 { "error", error },
             };
 
-            environment.AnalyticsClient.CaptureEvent("Chains", properties);
+            // todo log error using new event format
```
kalambet, line 52: Out of curiosity will this `todo` work in IDE warning system you need proper `TODO:`?

Arshia001 (changes requested):
This PR should be rebased on main after 453 is merged in.

src/UnityPackage/Assets/Web3Unity/Scripts/Prefabs/WebGL/WebGLBridge/Web3GL.cs
```diff
@@ -95,7 +93,8 @@ public static void AddTokenFunctionWeb(string tokenAddress,string tokenSymbol,st
             GasLimit = _gasLimit,
             GasPrice = _gasPrice
         };
-        await GameLogger.Log(PlayerPrefs.GetString("ChainId"), PlayerPrefs.GetString("RPC"), data);
+        // Temporarily disabled, waiting for migration to DI
```
kalambet, line 96: Maybe it makes sense to add `TODO:` here and on the next occasions too? I don't think VS or Rider supports `FIXME:`, though.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Evm/Dto/Chains.cs
Arshia001: It works in VS. No idea about Rider though.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/ApiAnalyticsClient.cs
Arshia001: If we're going to allow on-prem installation of the API server, then it definitely makes sense to make the API endpoint configurable. It can live next to the other configurations (chain ID etc.)

Arshia001: This complicates the analytics feature though, since to report the data to the correct backend, the events should be sent to our instance of the API server. And if that's the only API endpoint we're using in version 2, then maybe it doesn't make sense to make it configurable?

kalambet: Yeah, this is a good one. So far I see two reasons to make it configurable: the first reason is to separate staging and production environments because we will find ourselves in a mess of `Project IDs` that do not exist on pord while reporting analytics from stage; the second reason is to prepare for the self-hosted solution.

So I see there two solutions:
1. Make it configurable and make putting an empty URL in settings an act of opting out of the analytics collection.
2. Ask API to deal with both sets of Project IDs on the analytics endpoint and introduce a flag to opt-out of the analytics collection.

src/ChainSafe.GamingSdk.EVM.Unity/UnityEnvironmentExtensions.cs
oleksandrchainsafe: Should I add an XML comment here or do you mean web documentation?

src/ChainSafe.GamingSDK.EVM/Web3/Core/Evm/Dto/Chains.cs
oleksandrchainsafe: Yeah, it will in Rider. All `todo` no matter the case and `throw new NotImplementedException();` show up in Rider's todo tab.

src/ChainSafe.GamingSdk.EVM.Unity/UnityEnvironmentExtensions.cs
Arshia001: Web documentation. Just something to keep in mind.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/ApiAnalyticsClient.cs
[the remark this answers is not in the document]
oleksandrchainsafe: What would be the benefit of this change?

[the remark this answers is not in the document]
oleksandrchainsafe: I think it's just gonna create more coupling to a static class from different parts of code or even different modules, and that is no good.
We could go the way where IApiConfig is passed to the AnalyticsClient constructor as a dependency, I guess.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/ApiAnalyticsClient.cs
Arshia001: I've created a new issue for this: #468

Arshia001 (approved):

kalambet (approved):

PR #471, opened 2023-06-14
Update AnalyticsEvent.cs

KBryan:
Updated style of de-structuring accessing Key and Value properties of KeyValuePair directly

### DIFF
--- a/src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/AnalyticsEvent.cs
+++ b/src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/AnalyticsEvent.cs
@@ -41,9 +41,20 @@ namespace ChainSafe.GamingWeb3.Analytics
             IfNotNullNorWhitespace(GasLimit, () => AppendMember(nameof(GasLimit), GasLimit));
             IfNotNullNorWhitespace(GasPrice, () => AppendMember(nameof(GasPrice), GasPrice));
 
-            foreach (var (name, value) in CustomProperties)
+            if (CustomProperties != null)
             {
-                AppendMember(name, value);
+                foreach (var pair in CustomProperties)
+                {
+                    var key = pair.Key ?? "Unnamed";
+                    var value = pair.Value ?? "Unknown";
+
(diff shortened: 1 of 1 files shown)


src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/AnalyticsEvent.cs
```diff
@@ -41,9 +41,20 @@ public override string ToString()
...
-            foreach (var (name, value) in CustomProperties)
+            if (CustomProperties != null)
             {
-                AppendMember(name, value);
+                foreach (var pair in CustomProperties)
+                {
+                    var key = pair.Key ?? "Unnamed";
+                    var value = pair.Value ?? "Unknown";
+
+                    if (pair.Key == null || pair.Value == null)
```
RyRy79261, line 51: Wouldn't these values, if previously null, be assigned "Unnamed", meaning this would never trigger?

KBryan: Yes, I was wondering how to handle this. Do we want to log these errors or throw an exception?

KBryan: Yes, I'm aware of that going to remove the key/value set.

src/ChainSafe.GamingSDK.EVM/Web3/Core/Analytics/AnalyticsEvent.cs
```diff
@@ -41,9 +42,17 @@ public override string ToString()
...
 
-            foreach (var (name, value) in CustomProperties)
+            if (CustomProperties != null)
             {
-                AppendMember(name, value);
+                foreach (var pair in CustomProperties)
+                {
+                    if (pair.Key == null || pair.Value == null)
+                    {
+                        // TODO: Should integrate logging mechanism. Currently skips if key/value pair not found
```
RyRy79261, line 51: This doesnt currently skip if the pair is null

Arshia001 (approved):

PR #476, opened 2023-06-16
Run a non-unity restore after building the Unity project

Arshia001:


src/ChainSafe.GamingSdk.EVM.Unity/build.bat
```diff
@@ -1 +1,5 @@
-dotnet build /property:Unity=true
\ No newline at end of file
+dotnet build /property:Unity=true
+
+pushd ..\..
+dotnet restore
```
RyRy79261, line 4: Shouldn't this happen before the build, since this had to run to fix the errors in the build?

src/ChainSafe.GamingSdk.EVM.Unity/build.sh
```diff
@@ -1,2 +1,6 @@
 #! /usr/bin/env sh
-dotnet build /property:Unity=true
\ No newline at end of file
+dotnet build /property:Unity=true
+
+pushd ../..
+dotnet restore
```
RyRy79261, line 5: See `.bat` comment

src/ChainSafe.GamingSdk.EVM.Unity/build.bat
Arshia001: No, this serves to fix the state of the solution *after* the build, so you won't be getting the Nethereum errors you've been getting.

KBryan: Do we also need this in the process `dotnet build /property:Unity=True` ?  Or should that be a footnote and used manually based on the return to the editor?

Arshia001: Yes, that's what's actually doing the build.

oleksandrchainsafe (approved):

KBryan (approved):

src/ChainSafe.GamingSdk.EVM.Unity/build.bat
RyRy79261: What is the purpose of this file?

PR #479, opened 2023-06-19
Update README with building instructions

Arshia001:


KBryan (commented):
This looks good. In my workings, I needed to open the Unity project first and then run the publish to Unity script. I think we may also need the installation `com.unity.nuget.newtonsoft-json@3.0.2`

Arshia001:
> This looks good. In my workings, I needed to open the Unity project first and then run the publish to Unity script. I think we may also need the installation `com.unity.nuget.newtonsoft-json@3.0.2`

The newtonsoft-json package is pulled in automatically as a dependency of ours, so no need to add it anywhere, hurray 🎉

I'll take a look at whether we need to open Unity first, and update this accordingly if it's required. I don't think it should be required though.

KBryan (approved):
Looks good, as discussed there may be some oddities with different versions of Unity but the overall process seems to comply with what is needed to install successfully.

oleksandrchainsafe (approved):

PR #487, opened 2023-06-21
Add setup scripts

Arshia001:


setup.bat
```diff
@@ -0,0 +1,9 @@
+@echo off
+
+REM clone submodules
+git submodule update --init
+
+REM publish DLLs to unity package
+pushd src\ChainSafe.GamingSdk.EVM.Unity
```
oleksandrchainsafe, line 7: Don't get what pushd/popd commands do, but src\ChainSafe.GamingSdk.EVM.Unity must be updated after the "feature/fix-namespaces" branch is merged.

oleksandrchainsafe (approved):

setup.bat
Arshia001: @oleksandrchainsafe pushd is the same as cd, except it puts the previous directory on a stack and you can switch back using popd. I'll merge this in, so please update the path in the namespaces PR.

PR #522, opened 2023-07-12
Fix tests

Arshia001:
Merge after #486


src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
```diff
@@ -1,67 +1,93 @@
...
+            var balance = web3.RpcProvider.GetBalance(walletAddress).Result;
             var balanceFormatted = Units.FormatEther(balance);
             Assert.Greater(balanceFormatted, "0");
         }
 
         [Test]
         public void GetCodeTest()
         {
-            var code = infuraProvider.GetCode(goerliAddress).Result;
+            var code = web3.RpcProvider.GetCode(walletAddress).Result;
```
oleksandrchainsafe, line 81: What is this code?

Arshia001: No idea. I assume this is the RPC method we're calling: https://www.quicknode.com/docs/ethereum/eth_getCode

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
```diff
@@ -133,43 +159,61 @@ public void GetGasPriceTest()
...
             var receipt =
-                infuraProvider.GetTransactionReceipt(
+                web3.RpcProvider.GetTransactionReceipt(
                     firstTransactionInBlock.Hash).Result;
             Assert.AreEqual(firstTransactionInBlock.Hash, receipt.TransactionHash);
             Assert.AreEqual(firstTransactionInBlock.BlockHash, receipt.BlockHash);
         }
 
         /*
+         * To run this test you need to deploy a contract using your Ganache instance and the contract needs to have a safeMint function that increments the tokenId counter
```
oleksandrchainsafe, line 172: We should tag @juans-chainsafe there probably.

oleksandrchainsafe (approved):

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
juans-chainsafe: yeah, this one was the most difficult to migrate, not sure if we can automate it all because it requires a lot of manual steps

Arshia001: This is actually Juan's code. I do think it should be possible to use the SDK itself to deploy a smart contract in the setup code; but I haven't verified this.

RyRy79261 (commented):
Looks good though not sure how to run the code so cant approve

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
```diff
@@ -1,67 +1,93 @@
...
             */
+
+            // Deploy a contract with https://remix.ethereum.org/ and your Ganache instance.
+            // Make sure to select Ganache Provider as the Environment. Once deployed, replace the _contractAddress value with your contract address
+            // todo: This is impossible to automate
+            /* contractAddress = "0x2061c2B28F007DD0eD5064A61d352521CC1d2827"; */
```
mpetrunic, line 59: Let's not do this please. Use ganache cli to start ganache cli to start local blockchain and use nethereum to deploy contract before tests^^

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
mpetrunic: It's the compiled bytecode of the contract^^

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
```diff
@@ -1,67 +1,93 @@
...
         }
 
         [Test]
         public void GetLastBlockTest()
         {
-            var latestBlock = infuraProvider.GetBlock().Result;
+            var latestBlock = web3.RpcProvider.GetBlock().Result;
             Assert.AreEqual("0x0000000000000000", latestBlock.Nonce);
-            Assert.AreEqual("30000000", latestBlock.GasLimit.ToString());
+            Assert.AreEqual("6721975", latestBlock.GasLimit.ToString());
```
mpetrunic, line 90: how is this support to work?

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
mpetrunic: In any case you can use nethereum to deploy contract. It should be less than 5 loc

https://docs.nethereum.com/en/latest/contracts/deploying/

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
Arshia001: No idea. It always works on my machine, can't tell if it'd work with any Ganache instance.

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
Arshia001: I have never seen any good coming from starting CLI stuff in tests. However, the contract deployment step should indeed be automated.

mpetrunic: I don't know any other way to start local blockchain in C#. It's kinda common practice with blockchain to start dev node locally before tests and close it afterwards

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
mpetrunic: ah probably hardcoded in ganache instance

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
mpetrunic: I'm not saying that you have to run cli from c#, having some script that starts ganache, runs tests and then closes ganache is fine also^^

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
juans-chainsafe: I think those are the default values for a new ganache instance

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
Arshia001: I generally prefer including instructions with the tests: "A local Ganache node must be running on localhost:xxxx for these tests to run" It's also simpler to set up CI that way. The script would have to be able to locate the CLI, and you'd have to mess with PATH and stuff.

kalambet:
@juans-chainsafe and @oleksandrchainsafe tests are fixed and fully automated!

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
kalambet: Updated the test for EOA and SC addresses.

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
kalambet: The final solution is to run `anvil` for the test execution automation.

src/ChainSafe.GamingSDK.EVM.Tests/ChainsafeRPCTests.cs
kalambet: Default from the `anvil` "GoChain Testnet" now is used in this test,

src/ChainSafe.GamingSDK.EVM.Tests/ProvidersSendTests.cs
```diff
@@ -1,59 +1,72 @@
...
 
         [OneTimeSetUp]
         public void SetUp()
         {
-            web3 = Web3Util.CreateWeb3().Result;
-            secondAccount = Web3Util.CreateWeb3(1).Result;
+            node = Emulator.CreateInstance();
+
+            var firstAccountTask = CreateWeb3(0).AsTask();
+            firstAccountTask.Wait();
```
oleksandrchainsafe, line 29: Why not just request "Result" right ahead instead of "Wait()" here?

oleksandrchainsafe: We could also make this method (as well as all the test methods) async and use await then.

src/ChainSafe.GamingSDK.EVM/RPC/Contracts/Contract.cs
```diff
@@ -113,7 +113,9 @@ public async Task<object[]> Send(string method, object[] parameters = null, Tran
             if (txReq.GasPrice == null && txReq.MaxFeePerGas == null)
             {
                 var feeData = await provider.GetFeeData();
-                txReq.MaxFeePerGas = new HexBigInteger(feeData.MaxFeePerGas);
+
+                // txReq.MaxFeePerGas = new HexBigInteger(feeData.MaxFeePerGas);
```
oleksandrchainsafe, line 117: Is there a problem with `MaxFeePerGas`? Should we initialize GasPrice now instead?

oleksandrchainsafe (changes requested):

src/ChainSafe.GamingSDK.EVM.Tests/ProvidersSendTests.cs
kalambet: This is restriction of NUnit, unfortunately we can't do `OneTimeSetUp` and `OneTimeTearDown` `asnyc`.

src/ChainSafe.GamingSDK.EVM/RPC/Contracts/Contract.cs
kalambet: That was test to try to get `GasPrice`. Good point. I think it make sense to think about choosable strategy to tackle this.

Let me think how to properly refactor it.

src/ChainSafe.GamingSDK.EVM.Tests/ProvidersSendTests.cs
kalambet: Ah and overall it is a bad idea to request `Result` out of `ValueTask` since execution is not guarantied.

https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/
```
The short rule is this: with a ValueTask or a ValueTask<TResult>, you should either await it directly (optionally with .ConfigureAwait(false)) or call AsTask() on it directly, and then never use it again, e.g.
```

oleksandrchainsafe (approved):

juans-chainsafe (approved):
Looks good! I was able to run them locally as well.

This adds a lot of value

PR #523, opened 2023-07-14
Move Get* methods out of IRpcProvider and into an extension class

Arshia001:


src/ChainSafe.GamingSDK.EVM/Web3/Evm/JsonRpc/JsonRpcProvider.cs
```diff
@@ -68,26 +68,28 @@ public async Task<string[]> ListAccounts()
             }
         }
 
-        public async Task<TResponse> Send<TResponse>(string method, params object[] parameters)
+        public override async Task<T> Perform<T>(string method, params object[] parameters)
```
oleksandrchainsafe, line 71: I would keep TReponse (or TResult) as the type argument name so that it's more obvious what type you have to provide there. It's not important if T is only used as an argument type, but when it's for the return type and you always have to type it in manually, a little hint like TResult would be helpful.

oleksandrchainsafe (changes requested):

oleksandrchainsafe:
Overall that's awesome 🔥🔥🔥 Nice to see RpcProvider with just 4 members.

src/ChainSafe.GamingSDK.EVM/Web3/Evm/JsonRpc/JsonRpcProvider.cs
Arshia001: I'll do that in the event poller PR.

Arshia001:
@oleksandrchainsafe wait till you see the one in the event poller service branch then :D

juans-chainsafe (approved):

oleksandrchainsafe (approved):

Issue #531, opened 2023-07-25
iOS build not working


juans-chainsafe:
**Describe the bug**
iOS build is not working as expected, I get errors when I try to login in my device

**To Reproduce**
1. In Unity, go to Window - Web3Auth - Generate Deep Link and click **Generate** button
2. Switch the Build Settings to **IOS** platform
3. Click on **Build and Run** button
4. Select the folder to save the build
5. You should have a simulator in XCode or connected your real device
6. The app will be open in your device
7. Click **Login** button

**Expected behavior**
You should be prompted to connect to Metamask or Torus, depending what you have installed in your device/simulator

**Log errors**
```
InvalidOperationException: A suitable constructor for type 'Web3Unity.Scripts.Library.Ethers.Providers.JsonRpcProvider' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
  at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateConstructorCallSite (Microsoft.Extensions.DependencyInjection.ServiceLookup.ResultCache lifetime, System.Type serviceType, System.Type implementationType, Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteChain callSiteChain) [0x00000] in <00000000000000000000000000000000>:0 
  at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateExact (Microsoft.Extensions.DependencyInjection.ServiceDescriptor descriptor, System.Type serviceType, Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteChain callSiteChain, System.Int32 slot) [0x00000] in <00000000000000000000000000000000>:0 
  at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.TryCreateExact (System.Type serviceType, Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteChain callSiteChain) [0x00000] in <00000000000000000000000000000000>:0 
  at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateCallSite (System.Type serviceType, Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteChain callSiteChain) [0x00000] in <00000000000000000000000000000000>:0 
[...]

PR #546, opened 2023-08-08
Bugfix #531 iOS Constructor Missing

oleksandrchainsafe:
The problem was caused by Unity's stripping procedure.
Had to add **All our assemblies + Nethereum + NBitcoin + BouncyCastle** to the list of assemblies that will be ignored during the code stripping procedure.

This is done by adding a link.xml file to one of the projects and marking it as an embedded resource.

No matter what Stripping Level user chooses in his project settings, the code will still be there after the build is completed (Didn't test other levels though).


juans-chainsafe (approved):
iOS build worked correctly without changing any build setting. I used the login scene and the gelato examples to test.
But we need to retest this when we have the autogenerated unitypackage of the new version. It should work as well without a problem, but just in case.

kalambet (approved):

PR #557, opened 2023-08-15
393 make a single sign in scene with both metamask and web3auth

oleksandrchainsafe:


juans-chainsafe (changes requested):
@oleksandrchainsafe this is looking pretty good and organized!

I guess is still work in progress, but this is what I found:

- Wallet address is always 0
Screenshot 2023-08-16 at 16 20 45
- Logout button doesn't work
- The idea here is to add ERC721 and ERC1155 here?
Screenshot 2023-08-16 at 16 22 20

- In WebGL we may have an issue with the space on the screen 
Screenshot 2023-08-16 at 16 29 01

src/UnitySampleProject/Assets/Scenes/SampleMain.unity
```diff
@@ -0,0 +1,4033 @@
...
+      propertyPath: m_text
+      value: Logout
+      objectReference: {fileID: 0}
+    - target: {fileID: 794121698757186486, guid: ffa826e418b59974d83a0a74e316a03a, type: 3}
+      propertyPath: m_HorizontalAlignment
+      value: 2
+      objectReference: {fileID: 0}
+    - target: {fileID: 797592591053775153, guid: ffa826e418b59974d83a0a74e316a03a, type: 3}
+      propertyPath: m_Name
+      value: Button - Login
```
juans-chainsafe, line 3517: change to Button - Logout

Issue #535, opened 2023-07-27
Android build not working 


juans-chainsafe:
**Describe the bug**
Android build is not working as expected, I get errors when I try to login in my device

**To Reproduce**
1. Switch the Build Settings to **Android** platform
2. Click on **Build and Run** button
3. Select the folder to save the build
4. You should have a emulator in Android studio or connected your real device
5. See error in Unity

**Expected behavior**
Build should be successful in Unity

**Log errors**
```
Win32Exception: ApplicationName='/Applications/Unity/Hub/Editor/2021.3.22f1/PlaybackEngines/AndroidPlayer/SDK/tools/bin/sdkmanager', CommandLine='--list', CurrentDirectory='/Users/juanmanuelspoleti/Desktop/workspace/web3.unity/src/UnitySampleProject', Native error= Access denied
System.Diagnostics.Process.StartWithCreateProcess (System.Diagnostics.ProcessStartInfo startInfo) (at <92163ec09fd047358e34bc4648abe748>:0)
System.Diagnostics.Process.Start () (at <92163ec09fd047358e34bc4648abe748>:0)
(wrapper remoting-invoke-with-check) System.Diagnostics.Process.Start()
UnityEditor.Utils.Program.Start (System.EventHandler exitCallback) (at /Users/bokken/build/output/unity/unity/Editor/Mono/Utils/Program.cs:48)
UnityEditor.Utils.Program.Start () (at /Users/bokken/build/output/unity/unity/Editor/Mono/Utils/Program.cs:32)
UnityEditor.Android.Command.Run (System.Diagnostics.ProcessStartInfo psi, UnityEditor.Android.Command+WaitingForProcessToExit waitingForProcessToExit, System.String errorMsg) (at <d9aa0349546a4daeaefd100a44ff9f59>:0)
UnityEditor.Android.AndroidSDKTools.RunAndroidSdkTool (System.String toolName, System.String arguments, System.Boolean updateCommand, System.String errorMsg, System.String toolsDir, System.String[] warningsToIgnore) (at <d9aa0349546a4daeaefd100a44ff9f59>:0)
UnityEditor.Android.AndroidSDKTools.ListComponentsVersions () (at <d9aa0349546a4daeaefd100a44ff9f59>:0)
UnityEditor.Android.SDKManager.UpdatePackagesList () (at <d9aa0349546a4daeaefd100a44ff9f59>:0)
[...]

PR #564, opened 2023-08-23
#535 Android build broken

oleksandrchainsafe:
This PR fixes the Android build procedure.
There was a problem with `chainsafe.web3-unity.web3auth.Editor.asmdef` configuration inside `io.chainsafe.web3-unity.web3auth`.
Now it's gone.

The problem described in the task description is probably just an Android tools installation issue. So this PR actually fixes a different build bug than described in the ticket.


kalambet (approved):
LGTM

juans-chainsafe (approved):
Working! because I'm in Mac I needed to run this command: `sudo chown -R juanmanuelspoleti: /Applications/Unity` so the build doesn't fail, this is not necessary in windows.
Maybe we can add that line on the ChainSafe docs so people are aware of this, if they use mac.

And the UI doesn't look good in Android devices, but the scenes will change so, I'm not worry about this right now.
Screenshot 2023-08-23 at 11 26 07

PR #578, opened 2023-08-29
UI is more responsive

robGG1997:
- all UI elements are visible in mobile and free aspect ratio resolutions.
image
image
image
image
image
image

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,21 @@
+# Set to true to add reviewers to pull requests
+addReviewers: true
+
+# Set to true to add assignees to pull requests
+addAssignees: author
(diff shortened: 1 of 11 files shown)


sneakzttv (dismissed):

juans-chainsafe (approved):
UI it is much more responsive now in Android builds! Looks awesome.
I wasn't able to check it on iOS builds due to this issue https://github.com/ChainSafe/web3.unity/issues/581

If is fixed before merging this PR I will take a look, otherwise when iOS build is fixed I will check if it looks like Android build

sneakzttv (approved):

sneakzttv (approved):

PR #579, opened 2023-08-29
Removing resources folder creation

sneakzttv:
Removal of the resources folder being created so it doesn't create an empty folder seeing as we already have it there for web3auth now.


kalambet (approved):
LGTM

robGG1997 (approved):

PR #576, opened 2023-08-28
Resolved WebGl crash on SampleSceen

RyRy79261:
closes #553 

Blocked by PR #562

PR #582, opened 2023-08-30
Warning cleanup

sneakzttv:
Fixes nunit tests, foundry version was the culprit. GJ on finding it Peter.

Fixes the project id web request on minor code changes as having this is rather annoying.

Fixes a few annoying editor warnings regarding the absence of await in async calls.

3 login warnings:
2 calls were async without an await in login.cs
1 call was async without an await, removed async & returned a task completion instead in sha3behaviour.cs

Login process seems fine my end.

I have another PR coming in on top of this one after to fix the nft texture but I figured I'd wait until this is in to keep some separation for readability.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,6 +19,8 @@ jobs:
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly-ae3ec74234fe30a88279b3850c99ff191f373781
         
       - name: Setup .NET Core SDK ${{ matrix.dotnet-version }}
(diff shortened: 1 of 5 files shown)


kalambet (approved):
LGTM

juans-chainsafe (approved):
All looks working as expected, WebGL build is failing but is also failing in main so is not an issue from this PR.
I assume the WebGL issue will be fixed here https://github.com/ChainSafe/web3.unity/pull/576

sneakzttv:
sweet will just wait for @oleksandrchainsafe to give it the ok and then we'll merge if all good

Issue #498, opened 2023-06-27
ImportNFTTextureExample script not working correctly


juans-chainsafe:
**Describe the bug**
Error says that "Array cannot be null" in `Response data = JsonUtility.FromJson<Response>(System.Text.Encoding.UTF8.GetString(webRequest.downloadHandler.data));`

**To Reproduce**
1. Add an empty game object in the sample scene
2. Add the script 'ImportNFTTextureExample' in the game object
3. Run it with the Login scene

**Expected behavior**
The "Revised URI" is displayed in the console

**Screenshots**
<img width="686" alt="Screenshot 2023-06-27 at 13 59 01" src="https://github.com/ChainSafe/web3.unity/assets/104365141/6eac97e6-aff4-4c42-aac2-10ae5fedf327">

**Additional context**
**This issue was found while testing https://github.com/ChainSafe/web3.unity/pull/486**

```
ArgumentNullException: Array cannot be null.
Parameter name: bytes
System.Text.Encoding.GetString (System.Byte[] bytes) (at <b6c5d0f08ddf477186f780f49bb14dee>:0)
ImportNFTTextureExample.Start () (at Assets/Scripts/Samples/ERC1155/ImportNFTTextureExample.cs:27)
System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__7_0 (System.Object state) (at <b6c5d0f08ddf477186f780f49bb14dee>:0)
UnityEngine.UnitySynchronizationContext+WorkRequest.Invoke () (at /Users/bokken/build/output/unity/unity/Runtime/Export/Scripting/UnitySynchronizationContext.cs:153)
UnityEngine.UnitySynchronizationContext.Exec () (at /Users/bokken/build/output/unity/unity/Runtime/Export/Scripting/UnitySynchronizationContext.cs:83)
UnityEngine.UnitySynchronizationContext.ExecuteTasks () (at /Users/bokken/build/output/unity/unity/Runtime/Export/Scripting/UnitySynchronizationContext.cs:107)
```

Arshia001:
This error is happening because the relevant data cannot be fetched from `ipfs.io`. I've updated the code to report errors in the web request, but I have no idea why `ipfs.io` is not working. I can access the data using `ipfs.chainsafe.io`.

sneakzttv:
Fixed in the upcoming release with #587

PR #587, opened 2023-08-31
Texture-fix

### LINKED ISSUE #566: ImportNFTTexture sample is broken
Server returns error 422 when requesting the NFT metadata by NFT's URI. See Erc1155Sample.cs, ImportNftTexture method.

sneakzttv:
URI-call updated so it doesn't point to empty metadata.
Texture call updated and working. Closes #566 
Texture was deserializing to a class with the key "ImageUri", altered as this should be "image" following metadata standards.
Both scripts and prefab metadata for the above are updated also.
Another empty async call removed (left behind previously).
Unity engine import added geleto to script to enable webgl boot as Time.time was showing as missing a reference without it.

### DIFF
--- a/src/UnityPackages/io.chainsafe.web3-unity/Editor/Startup.cs
+++ b/src/UnityPackages/io.chainsafe.web3-unity/Editor/Startup.cs
@@ -49,7 +49,7 @@ namespace ChainSafe.GamingSdk.Editor
             };
         }
 
-        static async void ValidateProjectID()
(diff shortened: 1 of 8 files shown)


juans-chainsafe (approved):
Working as expected! awesome work.

This will also close https://github.com/ChainSafe/web3.unity/issues/498

kalambet (approved):
LGTM!

Issue #574, opened 2023-08-28
Some Sample Operations throw errors


robGG1997:
**Description**
Selective Sample Operations throw errors after logging in

**To Reproduce**
1. Login using an existing wallet
2. Click on any of the following operations. 
- Name & Symbol under ERC 20
- OwnerOf, OwnerOfBatch & All(Legacy Indexer) under ERC 721
3. An error is thrown

**Expected behavior**
Operations should complete without throwing an error, if there's missing input there should be error handling to display error on UI

**Desktop**
 - Chainsafe SDK version: 2.5
 - Unity Platform: Windows
 - OS: Windows

juans-chainsafe:
@robGG1997 can you add more detail on this? is working fine on my side using Goerli:
<img width="410" alt="Screenshot 2023-08-30 at 11 02 52" src="https://github.com/ChainSafe/web3.unity/assets/104365141/b7cd37a9-6312-4fa9-b4a3-00b89a3514b7">

The only one failing is All(Legacy Indexer) but I remember @kalambet mentioned that he is working on that.

PR #588, opened 2023-09-01
Sample operations error handling 574

robGG1997:
- In any case of an error we now display a message with the exception's message
- However I was able to resolve my errors for name and symbol sample operations with the help of @sneakzttv 
- The issue was a mismatch in network between my server settings, my login wallet address and contract address


juans-chainsafe (approved):
Working as expected, looks good in WebGL too!

Screenshot 2023-09-01 at 11 10 35

@robGG1997 we can close https://github.com/ChainSafe/web3.unity/issues/574 then, because was a configuration issue

sneakzttv (approved):
flame on!

PR #594, opened 2023-09-06
Automate sample modifications duplication/copying

robGG1997:
- this will automate the dev process of duplicating modifications to the imported package sample
- uses a bash script to remove and duplicate samples folder
- uses a git action workflow duplicate_package_sample.yml to check for changes to imported package sample to run duplicate_package_sample.sh and commit and push any changes if they exist
- when making a push with a modified imported package sample make sure to do a pull right after (give the action workflow a minute or two) if you plan on pushing more changes to that branch since remote and local git will have different HEADs


sneakzttv (approved):

duplicate_package_sample.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Specify the source directory and the destination directory
+SOURCE_DIRECTORY="src/UnitySampleProject/Assets/Samples/web3.unity SDK/2.5.0-pre001/Web3.Unity Samples/"
+DESTINATION_DIRECTORY="src/UnityPackages/io.chainsafe.web3-unity/Samples~/Web3.Unity/"
```
kalambet, line 5: Is it necessary to have `~` in the Samples folder name?

.github/workflows/duplicate_package_sample.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Run duplicate script file
+        run: bash ${GITHUB_WORKSPACE}/duplicate_package_sample.sh
+        
+      - name: Commit and Push
+        run: |
+          git config --global user.email "robel@chainsafe.io"
```
kalambet, line 19: Is it possible to add this to the pre-commit hook for example and now extra commit after?

duplicate_package_sample.sh
robGG1997: it's not necessary, it's used to exclude samples from the Unity hierarchy, because of the ~ samples can't be found/referenced from project hierarchy and can only be imported.

.github/workflows/duplicate_package_sample.yml
robGG1997: that could be better I have not tried, I'll check

PR #595, opened 2023-09-06
automated duplicating imported samples modifications

robGG1997:
this will automate the dev process of duplicating modifications to the imported package sample
uses a bash script to remove and duplicate samples folder
uses a git actions workflow duplicate_sample.yml to check for changes to imported package sample to run duplicate_package_sample.sh and commit and push any changes if they exist
This will only be automated to add a commit when a push happens to main (successful pull request)


RyRy79261:
@robGG1997 We might want a shell script for Windows as well

sneakzttv:
> @robGG1997 We might want a shell script for Windows as well

Just use git, it gives you bash for windows https://git-scm.com/downloads

sneakzttv (dismissed):

sneakzttv:
> @sneakzttv I normally use WSL2 so anything git related runs out of that section, but the issue is pathing buggers out for all the .Net stuff for me, so I just run directly out of windows

Fair

sneakzttv:
hmm...added a comment, initialized the build then reimported sample via the package manager and the comment had been reset.

robGG1997:
> hmm... forked the branch, added a comment and made a commit. Initialized the build then reimported sample via the package manager and the comment had been reset. If this is because the branch was forked then ignore me, otherwise it was working fine on initial import. @robGG1997 just confirm when you can and i'll approve based on said outcome.

Have you set up pre-commit in .git/hooks?

robGG1997:
> @robGG1997 We might want a shell script for Windows as well

based on the new changes the bash script is going to be used by a GitHub action on a push to main, no need to use it manually

robGG1997:
> @robGG1997 We might want a shell script for Windows as well

based on the new changes the bash script is going to be used by a GitHub action on a push to main, no need to use it manually

sneakzttv:
so...we wait?

robGG1997:
> so...we wait?

When @kalambet approves this we're good to go

kalambet:
I'm still concern a bit with extra commit. There is no urgency in solving this so far, however I'll try to finalize my thoughts till the end of the day. Sorry :(

.github/workflows/duplicate_sample.yaml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v2
+
+      - name: Run duplicate script file
+        run: bash ${GITHUB_WORKSPACE}/duplicate_package_sample.sh
+
+      - name: Commit and Push
+        run: |
+          git config --global user.email "robel@chainsafe.io"
+          git config --global user.name "Robel Getnet"
+          git config --global user.username "robGG1997"
```
kalambet, line 24:
```suggestion
          git config --global user.email "${{ github.actor }}@users.noreply.github.com"
          git config --global user.name "${{ github.actor }}"
```

kalambet: I think there is no better solution right now. 
However let's personalize commit to the person who initiated workflow to run.

robGG1997: done, now the commit author will be whoever initiated the workflow, I've also tested this

kalambet (approved):
LGTM

kalambet:
@robGG1997 Please update the branch with the changes from `main`

sneakzttv (approved):

kantagara (approved):

PR #624, opened 2023-09-18
Multicall V2.5 implementation 

### LINKED ISSUE #484: Implement multicall on the SDK side
There is an existing implementation in the API, which can be moved to the SDK side. This is the code in question:

https://github.com/ChainSafe/game-api/blob/main/src/routes/evm/multicall.ts

RyRy79261:
closes #484


src/ChainSafe.Gaming/MultiCall/IMultiCallConfig.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace ChainSafe.Gaming.MultiCall
+{
+    public interface IMultiCallConfig
```
robGG1997, line 5: is this useful? it's not referenced anywhere and we use the concrete config class for DI

src/ChainSafe.Gaming/MultiCall/MultiCallCommonMethods.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace ChainSafe.Gaming.MultiCall
+{
+    public static class MultiCallCommonMethods
+    {
+        public const string Aggregate = "aggregate";
+        public const string Aggregate3 = "aggregate3";
+        public const string Aggregate3Value = "aggregate3Value";
+        public const string BlockAndAggregate = "blockAndAggregate";
+        public const string GetBasefee = "getBasefee";
```
robGG1997, line 9: small typo fix, this isn't in camelCase, `GetBaseFee`

src/ChainSafe.Gaming/MultiCall/MultiCall.cs
```diff
@@ -0,0 +1,114 @@
...
+            if (chainConfig.ChainId != null && MultiCallDefaults.DeployedNetworks.Contains(chainConfig.ChainId))
+            {
+                multiCallContract = builder.Build(ContractAbi, DefaultDeploymentAddress);
+            }
+            else
+            {
+                if (chainConfig.ChainId != null && config.CustomNetworks.TryGetValue(chainConfig.ChainId, out var address))
+                {
+                    multiCallContract = builder.Build(ContractAbi, address);
+                }
```
robGG1997, line 33: should we maybe throw an exception for an `else` here?

src/ChainSafe.Gaming/MultiCall/MultiCall.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        public async Task<List<Result>> MultiCallAsync(Call3Value[] multiCalls, int pageSize = DefaultCallsPerRequest)
+        {
+             if (multiCalls.Any(x => x.Value > 0))
```
robGG1997, line 39: can't this `if/else` be moved in a single foreach loop to avoid nesting and optimize code? I see a lot of common statements in both paths/conditions.

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Samples/MultiCallSample.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            var decodedTotalSupply = erc20Contract.Decode(CommonMethod.TotalSupply, multicallResultResponse[1].ReturnData.ToHex());
+            Debug.Log($"decodedTotalSupply {((BigInteger)decodedTotalSupply[0]).ToString()}");
+        }
+
+    }
+
+    private static class CommonMethod
+    {
+        public const string BalanceOf = "balanceOf";
```
robGG1997, line 72: can't we just make these class constants? since they're not being referenced anywhere else

robGG1997: or we can move it to `Chainsafe.Gamming` and make it public and reference it from there

robGG1997 (changes requested):

Packages/io.chainsafe.web3-unity/Samples~/Web3.Unity/Scripts/Prefabs/MultiCall/MultiCall.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
```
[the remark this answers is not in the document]
RyRy79261: Ahh will nuke it

src/ChainSafe.Gaming/MultiCall/IMultiCallConfig.cs
RyRy79261: @robGG1997 The config interface is needed for the dependency injection but the config could later on be expanded to additional features, I left having the option of a config object so that the work flow is available down the line

src/ChainSafe.Gaming/MultiCall/MultiCall.cs
RyRy79261: Roger that

src/ChainSafe.Gaming/MultiCall/IMultiCallConfig.cs
robGG1997: I believe DI uses the concrete class not the interface, however I see your point we could expand on it in the future and it doesn't hurt to keep it

src/UnitySampleProject/Assets/Samples/web3.unity SDK/2.5.0/Web3.Unity Samples/Scripts/Prefabs/MultiCall/MultiCall.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    // Update is called once per frame
+    void Update()
+    {
+
+    }
+
+    public async void FetchBatchBalances()
+    {
+
```
kantagara, line 21: Is this still empty?

src/ChainSafe.Gaming/MultiCall/MultiCall.cs
robGG1997: What if the method body was something like this
```
var results = new List<object[]>();

            foreach (var page in multiCalls.Batch(pageSize))
            {
                var contractCalls = new List<Call3Value>();

                foreach (var multiCall in page)
                {
                    Call3Value call3Value = new Call3Value
                    {
                        CallData = multiCall.CallData, Target = multiCall.Target, AllowFailure = multiCall.AllowFailure,
                    };

                    if (multiCalls.Any(x => x.Value > 0))
                    {
                        call3Value.Value = multiCall.Value;
                    }

                    contractCalls.Add(call3Value);
                }

                if (contractCalls.Count <= 0)
                {
                    continue;
                }

                var callParams = new object[]
                {
                    contractCalls,
                };

                var callResults = await multiCallContract.Call(MultiCallCommonMethods.Aggregate3, callParams);

                results.Add(callResults);
            }

            return ExtractResults(results);
```

src/ChainSafe.Gaming/MultiCall/MultiCall.cs
robGG1997: Has this been updated? I still see it in changes

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Samples/MultiCallSample.cs
robGG1997: I'm not sure either, @oleksandrchainsafe where do you think this fits in?

Packages/io.chainsafe.web3-unity/Samples~/Web3.Unity/Scripts/Prefabs/MultiCall/MultiCall.cs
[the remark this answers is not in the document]
RyRy79261: Has been removed on my side, not sure why its not picked up

src/ChainSafe.Gaming/MultiCall/MultiCall.cs
RyRy79261: @robGG1997 all of the items need to be one type or the other if any are found to be a Call3Value type or a regular Call3 type so there cant be a mix

sneakzttv (approved):

kantagara (approved):

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Samples/MultiCallSample.cs
kantagara: @robGG1997 I'm not Oleksandr I think it makes sense for this to be in here because only multicall sample will ever need this, it makes sense that this is part of the multicallsample as a subclass.

robGG1997 (approved):

PR #634, opened 2023-09-26
Ramp.Network - FIAT ON/OFF RAMP for iOS 

kantagara:
It took some time to be delivered but it's here! 

**RampData**
This is a scriptable object that contains all the essential data that's needed for starting up the ramp. i.e. configuration from https://docs.ramp.network/mobile/ios-sdk this site

I have made a base abstract class
**RampIntegrationBase**
It has an abstract method called OpenRamp which is kind of a proxy, because based on the platform (iOS, Android) we'd have different implementations. Since it was merely impossible to pass all the parameters as a proper class, I've passed them in as individual parameters, it looks ugly, but our users shouldn't worry about it. They just need to call "OpenRamp" and that method should call the native one that we have linked for each platform (currently still iOS, android is coming).

It also has two Actions (since I couldn't make it a proper event). OnRampPurchaseEvent and OffRampSaleEvent. They're static because I had to have a static method inside of the iOS implementation otherwise we wouldn't be able to execute it at all. 

**RampIntegrationSample**
This is a simple script that our developers should actually only be 'concerned' about. 
 Because this is how they should be using our Ramp integration, so it's dead easy and straight forward. They should only be concered about those two events, and that's it. 

**RampIntegrationiOS**
iOS specific implementation of the RampIntegartion. It's irrelevant for our users tbh but it's nice for you guys to see it. 

There is also a SendCryptoPayload method in swift, but I don't think that's something we'd need? If it's needed, do let me know.


Packages/io.chainsafe.web3-unity/Samples~/Web3.Unity/Scripts/RampIntegration/RampIntegrationSample.cs
```diff
@@ -0,0 +1,41 @@
...
+public class RampIntegrationSample : MonoBehaviour
+{
+    [SerializeField] private RampData rampData;
+    private RampChainsafeIntegrationBase _ramp;
+
+    private void Awake()
+    {
+#if UNITY_IOS
+        _ramp = new RampChainsafeIntegrationiOS(rampData);
+#endif
```
oleksandrchainsafe, line 15: So while this looks simple & clean, I would move the Ramp object reference to the Web3 infrastructure. This way it would utilize dependency injection (we could use the current player address for example) and it would be accessible just like any other component provided by the SDK. Kinda like this:
`web3.RampExchanger().Buy(...)` or `web3.RampExchanger().Open(...)` or `web3.RampExchanger().Show(...)` idk
_See src/ChainSafe.Gaming.Gelato/GelatoExtensions.cs_

Choosing the required implementation can also be done silently so that users won't have to care about it.

oleksandrchainsafe: You would have to provide RampData to OpenRamp method, because there will only be one instance of RampExchanger.

src/UnitySampleProject/Assets/Plugins/iOS/RampChainsafeIntegration.framework/_CodeSignature/CodeResources
```diff
@@ -0,0 +1,168 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
oleksandrchainsafe, line 1: Package plugins should go into the package Plugins folder, not into the UnitySampleProject.

oleksandrchainsafe (changes requested):

src/UnitySampleProject/Assets/Plugins/iOS/RampChainsafeIntegration.framework/_CodeSignature/CodeResources
oleksandrchainsafe: Also, we could create a separate package for Ramp, as it is fully on the Unity project side.
(We haven't figured out how to break projects from the core solution into separate packages yet, but the idea is to keep each feature in a separate package)

kantagara: Yeah, I'd probably make those for the entire Fiat on/off ramp (like that should be a separate package), then people can choose their own fiat provider.

Packages/io.chainsafe.web3-unity/Samples~/Web3.Unity/Scripts/RampIntegration/RampIntegrationSample.cs
kantagara: My biggest concern in doing that is -> Can we call native iOS code with extern trough our gaming sdk?

oleksandrchainsafe: Do we have to?

kantagara: Unfortunately yes. The essential part here is the "MonoPInvokeCallback" in the RampExchangeriOS class that's actually a callback that is being called by the native iOS code.

kantagara: I did update the project structure tho. I'm not doing the creation of the RampExchanger subclasses in the Sample class, but I 've made the factory that will do that for each individual platform.

oleksandrchainsafe: Sorry, I don't understand. Aren't you already doing it for iOS?

kantagara: Yes, so you should update your webGL branch 😁

kantagara:
Everything is being done on Oleksandrs PR, so no need for this PR to be here at all anymore.

PR #674, opened 2023-10-19
Docstring for Web3 folder in Chainsafe.Gaming

oleksandrchainsafe:
_plus some small refactoring_

Some of the stuff was generated using ChatGPT while the other part was written by me. So if you find something is not descriptive enough, tell me.


src/ChainSafe.Gaming/Web3/Core/Web3.cs
```diff
@@ -12,8 +12,9 @@ namespace ChainSafe.Gaming.Web3
 {
     /// <summary>
     /// Facade for all Web3-related services.
+    /// Use this as an entry point to all SDK features.
     /// </summary>
-    public class Web3 : IAsyncDisposable
+    public class Web3 : IAsyncDisposable, IInitializableWeb3
```
robGG1997, line 17: Is this abstraction (`IIinitializableWeb3`) necessary? Since we're always using the concrete class we make the same calls from just web3, If there needs to be an abstraction I believe `IWeb3` makes sense instead `IInitializableWeb3` since we only have one concrete web3 type or do we plan on having more than one?

robGG1997 (changes requested):

src/ChainSafe.Gaming/Web3/Core/Build/Web3Builder.cs
```diff
@@ -63,14 +70,19 @@ public Web3Builder Configure(ConfigureServicesDelegate configureMethod)
             return this;
         }
 
+        /// <summary>
+        /// Build <see cref="Web3"/> object using the settings provided by this Web3Builder object.
+        /// </summary>
+        /// <returns><see cref="Web3"/> object.</returns>
         public async ValueTask<Web3> BuildAsync()
```
robGG1997, line 77: Building web3 is failing on Unity with runtime error `A suitable constructor for type 'ChainSafe.Gaming.Evm.ChainRegistryProvider' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor`

robGG1997 (dismissed):

src/ChainSafe.Gaming/Web3/Core/Web3.cs
oleksandrchainsafe: It's there just to hide the `InitializeAsync' method as it should only be called by the `Web3Builder`. It also has `internal` access modifier.

src/ChainSafe.Gaming/Web3/Core/Build/Web3Builder.cs
oleksandrchainsafe: Fixed

src/ChainSafe.Gaming/Web3/Core/Web3.cs
robGG1997: If that's the only reason maybe we should just change the accessor on the method, `internal InitializeAsync` should hide it on Unity.

robGG1997: Extra base types can be confusing if there's only one inheriting type, if we plan on adding more Web3 Types like we have signer and txn executor types this is fine but if there'll only be just one concrete type it's a bit confusing, especially with an -able suffix like _initializable_ which makes me think is there a non-initializable type 😅

oleksandrchainsafe: > If that's the only reason maybe we should just change the accessor on the method, `internal InitializeAsync` should hide it on Unity.

You're right. Why didn't I think of that?

oleksandrchainsafe: Fixed

robGG1997: Have you pushed the commit? I don't see any changes.

kalambet: Agree here on Initializable/Non-initializable points.

kalambet (changes requested):

src/ChainSafe.Gaming/Web3/Core/Web3.cs
oleksandrchainsafe: Snap. My git software didn't push and didn't show me a msg box either because I was not logged in.
It's pushed now.

oleksandrchainsafe: For me personally, having a Zenject background, IInitializable means that the class has the `Initialize` method. So I don't agree with Initializable/Non-initializable points :D

src/ChainSafe.Gaming/Web3/Core/Build/Web3ServiceCollectionExtensions.cs
```diff
@@ -26,6 +36,12 @@ public static void AssertConfigurationNotBound<T>(this IWeb3ServiceCollection se
             }
         }
 
+        /// <summary>
+        /// Register the specified implementation using 2 contract types.
+        /// </summary>
+        /// <typeparam name="TInterface1">First contract type.</typeparam>
+        /// <typeparam name="TInterface2">Second contract type.</typeparam>
+        /// <typeparam name="TImplementation">Implementation type.</typeparam>
```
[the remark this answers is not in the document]
oleksandrchainsafe: Thought we don't have to describe parameters with these modifiers as users don't pass them explicitly. Will add them now.

src/ChainSafe.Gaming/Web3/Core/Analytics/NoOpAnalyticsClientExtensions.cs
```diff
@@ -6,10 +7,21 @@ namespace ChainSafe.Gaming.Web3.Analytics
...
+        /// Disables analytics for the <see cref="Web3"/> instance.
+        /// </summary>
+        /// <returns>Service collection to enable fluent syntax.</returns>
         public static IWeb3ServiceCollection DisableAnalytics(this IWeb3ServiceCollection serviceCollection)
         {
             serviceCollection.Replace(ServiceDescriptor.Singleton<IAnalyticsClient, NoOpAnalyticsClient>());
             return serviceCollection;
         }
+
+        public static bool AnalyticsDisabled(this IWeb3ServiceCollection serviceCollection)
```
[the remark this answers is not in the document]
oleksandrchainsafe: Added

src/ChainSafe.Gaming/Web3/Core/Build/Web3Builder.cs
```diff
@@ -27,7 +27,6 @@ public Web3Builder()
                 .AddSingleton<IContractBuilder, ContractBuilder>();
         }
 
-        // TODO: inline parameterless constructor into this one (therefore remove that overload)
         public Web3Builder(IProjectConfig projectConfig, IChainConfig chainConfig)
```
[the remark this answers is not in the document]
oleksandrchainsafe: Added

src/ChainSafe.Gaming/Web3/Core/Build/Web3ServiceCollectionExtensions.cs
```diff
@@ -53,6 +77,13 @@ public static void AssertConfigurationNotBound<T>(this IWeb3ServiceCollection se
             serviceCollection.AddSingleton<TInterface2, TImplementation>(sp => sp.GetRequiredService<TImplementation>());
         }
 
+        /// <summary>
+        /// Register the specified implementation using 3 contract types.
+        /// </summary>
+        /// <typeparam name="TInterface1">First contract type.</typeparam>
+        /// <typeparam name="TInterface2">Second contract type.</typeparam>
+        /// <typeparam name="TInterface3">Third contract type.</typeparam>
+        /// <typeparam name="TImplementation">Implementation type.</typeparam>
```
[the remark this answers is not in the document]
oleksandrchainsafe: Added

sneakzttv (approved):
yay comments

kalambet (approved):

PR #676, opened 2023-10-19
Add docstrings to the RPC folder

### LINKED ISSUE #667: Write Docsstring for RPC folder Chainsafe.Gaming

kalambet:
Closes #667


RyRy79261 (approved):
Looks good, though I'm not sure if we might want to expand on certain definitions, like in the context of the `Contract` class, where we could say more for jargon heavy topics like, ABI, Calldata, Gas, Estimates, etc

kalambet:
@RyRy79261 I agree with you completely. The only reason I hesitated is that we still do not have clear strategy on the general purpose smart contracts. I think that can be improved as soon as we start tackling this part.

sneakzttv (approved):
more docstrings yes!

Issue #688, opened 2023-10-26
Can't interact with elements in iOS


juans-chainsafe:
**Describe the bug**
in a iOS build after the build finishes and the app opens in the device, you can't interact with buttons or do any actions

**To Reproduce**
1. Switch platform to iOS
2. Click on Build and Run
3. Wait for build to finish in Unity
4. Switch your development profile in Xcode
5. Start the build in Xcode selecting a real device or simulator
6. Wait for the app to start in your device
7. Try to click any button in the SampleLogin

**Expected behavior**
User can interact with the interface

**Mobile (please complete the following information):**
 - Chainsafe SDK version: [main]
 - Platform: [iOS]
 - Type: [Real device]
 - iOS version [17.0.3]

**Additional context**
`JsonSerializationException: Unable to find a constructor to use for type ChainSafe.Gaming.WalletConnect.Model`

sneakzttv:
Closed via https://github.com/ChainSafe/web3.unity/pull/689

PR #689, opened 2023-10-26
Fixed iOS buttons for web3 auth not working

kantagara:
Had to make sure that the entire walletconnect methods are there since Unity REALLY likes to strip down a lot of code 😄


juans-chainsafe (changes requested):
Error was fixed! but I can't login to the SampleMain, let me know if you want to fix this in a new PR:

When I try with existing wallet (I have MetaMask installed):
```
[Web3] Core already initialized
Stacktrace is not supported on this platform.

[Web3] [Relayer] Checking for established connection
Stacktrace is not supported on this platform.

[Web3] [Relayer] Restarting transport
Stacktrace is not supported on this platform.

[Web3] [Relayer] Sending request through provider
Stacktrace is not supported on this platform.

[Web3] [JsonRpcProvider] Checking if connected
Stacktrace is not supported on this platform.
```

- With Web3Auth I tried 3 times, the first 2 failed, but the 3rd worked.

- Also, Logout is not working for me in Web3Auth, I see in the console:
```
System.ArgumentException: Scalar is not in the interval [1, n - 1]
Parameter name: d
```

Edit: I mean to approve the changes, in case that you want to fix this in other PR, I can create it

juans-chainsafe (approved):

src/ChainSafe.Gaming.Unity/link.xml
```diff
@@ -19,10 +19,7 @@
     <assembly fullname="ChainSafe.Gaming.InProcessSigner"/>
     <assembly fullname="ChainSafe.Gaming.InProcessTransactionExecutor"/>
     <assembly fullname="ChainSafe.Gaming.NetCore"/>
-    <assembly fullname="ChainSafe.Gaming.WalletConnect">
```
robGG1997, line 22: Should we Preserve the entire assembly? We only need it for the WC JsonRPC methods, or is this causing problems?

kantagara: Yeah that was causing the problem unfortunately, because some parts of the WalletConnect.Model were also being stripped. So I wanted to be better safe than sorry and preserved everything.

robGG1997 (approved):

robGG1997:
LGTM!

kantagara:
Closes #688

kantagara:
@juans-chainsafe @robGG1997 Please re-review this, I mean, it works :D but I just updated from main and it's asking for new review again.

robGG1997 (approved):

sneakzttv (approved):

PR #694, opened 2023-10-31
Fixing the issue caused by web3auth not working due to it's using old version of BouncyCastle.Crypto.dll

kantagara:
Essentially had to modify the publish-to-unity-package.sh/bat so that we backup the current working version.
Wallet Connect works on WebGL.


RyRy79261 (approved):
Tested in editor and it worked on my end

juans-chainsafe (approved):
Working now in the editor and in a WebGL build!

Issue #660, opened 2023-10-16
Prepare Lootboxes for openUPM package submission


kantagara:
There are some small things that need to be tested properly before we submit the package to the open UPM.

- [ ] Change namespace from chainlink.lootboxes to lootboxes.chainlink
- [ ] Test out if you can actually import lootboxes and web3 package by itself to a brand new project so that lootboxes just work

sneakzttv:
picking this up!

sneakzttv:
Namespaces changed and it's importing without errors! Yeet yeet!
We still need to put it all on chain as it's currently connecting and running only locally.
Will wait for Oleksander to return to handle that.
https://github.com/ChainSafe/web3.unity/pull/700

PR #700, opened 2023-11-02
Lootbox namespace change and error checks

sneakzttv:
Issue:
We need to start preparing lootboxes for OPM submission, the scope of this task is a namespace change and error checks #660 

Updates:
- Namespace changed from chainlink.lootboxes to lootboxes.chainlink and all occurrences updated.
- Updated the transition in login.cs so that the lootbox scene may be used with the login scene if the user hasn't got the main scene set as their 2nd scene (we'll need this once the contracts are up on chain) [REVERTED]
- We still need to swap all the data so it can use the login scene to be connected to and to pull data from testnet/mainnet as opposed to a local chain, will wait for @oleksandrchainsafe regarding this before we submit to OPM.

To Test:
Import the lootbox package and run it in the editor. You don't need to use the login scene currently as the launcher script initializes from a local rpc node. Just update your web3config class variables in web3.unity\src\UnitySampleProject\Assets\Samples\web3.unity SDK Lootboxes\1.0.0\Chainlink Lootbox Sample\Scripts\ChainlinkLootboxSampleLauncher.cs to match your local node config.

Expected Behavior:
No errors on import due to the namespace changes.
If you're testing with a local RPC node, everything should work as intended.


kantagara (approved):

Packages/io.chainsafe.web3-unity/Samples~/Web3.Unity/Scripts/Scenes/Login.cs
```diff
@@ -56,7 +56,15 @@ protected async Task TryLogin()
 
             LoginSceneIndex = SceneManager.GetActiveScene().buildIndex;
             
-            SceneManager.LoadScene(MainSceneName);
+            // Checks if the user has removed the main scene from the build settings, allows for other packages like loot boxes to be tested with the login scene
```
robGG1997, line 59: We should avoid using scene build indexes, that's why I used name to find login scene, If we have web3auth imported this will end up loading the Login scene again so lets revert this

robGG1997 (changes requested):

Packages/io.chainsafe.web3-unity/Samples~/Web3.Unity/Scripts/Scenes/Login.cs
sneakzttv: reverted

RyRy79261 (approved):

PR #715, opened 2023-11-08
Owner of 721 and 1155 fix

### LINKED ISSUE #701: OwnerOf 721 and 1155 type error with integers
Owner of 721 and 1155 methods work fine with a string as the token id "0x23123123" etc but the call fails when using an integer. We should be inputting this as a string and then trying to parse it as an int if it doesn't start with "0x".

You can see a working example below on adil mainnet with a modified owner of method that takes an integer as a parameter, if the contract data builds the token id as a string it throws an error saying the token doesn't exist. We need to account for both cases here.

```
using System.Threading.Tasks;
using ChainSafe.Gaming.UnityPackage;
using UnityEngine;
using Web3Unity.Scripts.Prefabs;
using ABI = ChainSafe.Gaming.UnityPackage.ABI;

/* This prefab script should be copied & placed on the root of an object.
Change the class name, variables and add any additional changes at the end of the execute function.
The initialize function should be called by a method of your choosing */

/// <summary>
/// Fetches the owner of an ERC721 token id
/// </summary>
public class OwnerOfTest : MonoBehaviour
{
    // Variables
    private string contractAddress = "0xcb775d2E24eFE447379Fd6249870922A327dec41";
    private int tokenId = 431464;
    private string _abi = ABI.ERC_721;

    /// <summary>
    /// Starts the task, you can put this in the start function or call it from a button/event
    /// </summary>
    public async void InitializeTask()
    {
        await ExecuteTask();
    }

    /// <summary>
    /// Executes the prefab task and sends the result to the console, you can also save this into a variable for later use
    /// </summary>
    private async Task ExecuteTask()
    {
        var balance = await OwnerOf(contractAddress, tokenId);
        SampleOutputUtil.PrintResult(balance.ToString(), nameof(Erc721Sample), nameof(Erc721Sample.OwnerOf));
    }
    
    /// <summary>
    /// Modified owner of call to use an int instead of a string
    /// </summary>
    /// <param name="contractAddress"></param>
    /// <param name="tokenId"></param>
    /// <returns></returns>
    private async Task<string> OwnerOf(string contractAddress, int tokenId)
    {
        var method = "ownerOf";
        var contract = Web3Accessor.Web3.ContractBuilder.Build(_abi, contractAddress);
        var contractData = await contract.Call(method, new object[]
        {
            tokenId
        });
        return contractData[0].ToString();
    }
}
```

Proposed fix:
Add a check to see if the string begins with "0x", if not parse it into an integer and place the result into the contract call object.

robGG1997:
closes #701 

There is no `ownerOf` function for `ERC1155` because a token is not necessarily unique, it can have multiple owners. So I didn't implement it.
`BatchOwnerOf` already works with both types for `tokenId` so no need for a refactor there.


Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Token/Erc721.cs
```diff
@@ -36,13 +36,28 @@ public static async Task<int> BalanceOf(Web3 web3, string contractAddress, strin
...
+        }
+        
+        /// <summary>
+        /// Owner Of ERC721 Token
+        /// </summary>
+        /// <param name="web3"></param>
+        /// <param name="contractAddress"></param>
+        /// <param name="tokenId"></param>
+        /// <returns></returns>
+        public static async Task<string> OwnerOf(Web3 web3, string contractAddress, int tokenId)
```
kantagara, line 50: Are we 100% sure that the tokenId will never exceed ints maxvalue? Should it be biginteger maybe instead?

RyRy79261:
@robGG1997 

So with ERC1155, theres a concept of `BalanceOf` & `BalanceOfBatch` 

This allows you to see how many of a token, by its ID, an account has. If you have set up the token by that ID to be a singular, unique, NFT. Then it would return a balance of 1. 

If it is an NFT that you've issued a few of, say a badge for attending an event, multiple accounts could have a balance of 1.

If its gold in a game, then the balance would return more than 1 in some instances:
https://docs.openzeppelin.com/contracts/3.x/erc1155

Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Token/Erc721.cs
robGG1997: I'm not sure but sounds valid, thanks Nikola will update

robGG1997:
> @robGG1997
> 
> So with ERC1155, theres a concept of `BalanceOf` & `BalanceOfBatch`
> 
> This allows you to see how many of a token, by its ID, an account has. If you have set up the token by that ID to be a singular, unique, NFT. Then it would return a balance of 1.
> 
> If it is an NFT that you've issued a few of, say a badge for attending an event, multiple accounts could have a balance of 1.
> 
> If its gold in a game, then the balance would return more than 1 in some instances: https://docs.openzeppelin.com/contracts/3.x/erc1155

Understood, so I should refactor Balance function instead?

Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Token/Erc721.cs
RyRy79261: So an example of when this could become an issue is, as example, the original ChainSafe ERC1155 NFT contract.

We encode the IPFS media identifier (A hash to look up a file on distributed storage), we encoded the media hash, into it's tokenId.

This means the tokenId = the unique media hash. The issue is, as a number, it is extremely big, and would likely overflow most `Int` types.

So personally, I would prefer to not even entertain the idea that we would use `Int` at all except for smaller Solitity types, like `Uint128` or below. But thats a whole lot of edge case checking


For reference, the max `Uint256` as an integer
```
// uint256 MAX_INT = uint256(-1)
uint256 MAX_INT = 115792089237316195423570985008687907853269984665640564039457584007913129639935
```

kalambet: BigInteger for sure!

RyRy79261:
@robGG1997 Yeah I rate just adding the non batch variant for 1155 if its missing is fine. I still need to actually review the code though.

Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Token/Erc721.cs
robGG1997: Updated

robGG1997:
> @robGG1997 Yeah I rate just adding the non batch variant for 1155 if its missing is fine. I still need to actually review the code though.

Updated `BalanceOf` for `Erc1155`

sneakzttv (approved):

kantagara (approved):
LGTM!

juans-chainsafe (approved):
Samples working! tested with wallet connect and web3auth.

Also fixed the issue with ImportNftTexture scene ✅

Issue #680, opened 2023-10-23
Technical Debt Collection


boorich:
What to do?
Review the following tasks and tick the resolved ones so that we can remove them. When you are done, please add any domain-specific knowledge you might have about the remaining tasks either as a comment or directly within the list:

- [x] Implement callback functionality instead of polling for WebGLWallet. See: [Unity Forum Discussion](https://forum.unity.com/threads/c-jslib-2-way-communication.323629/#post-2100593)
- [x] Update WebPageWallet to use deep links instead of the clipboard (Note: WebPageWallet has been removed)
- [x] Close the web page when a copy/deep link is activated (Note: WebPageWallet has been removed)
- [x] Remove the warning or hint when using WebPageWallet and the user has no wallet installed (Note: WebPageWallet has been removed)
- [x] Change the background color of the WebPageWallet page to Chainsafe's gray (Note: WebPageWallet has been removed)
- [x] Adjust the behavior of ChainsafeServerSettings and WebGLTemplatesInstallerTool
- [x] Eliminate duplicated logic in the Chains class, considering the functionality of ChainProvider
- [x] Prompt users to install WebGLTemplates only when the current platform is WebGL
- [x] Incorporate the "Packages" project (Unity scripts outside the core solution) into the solution for better support during refactoring
- [x] Enable Unity solution updates without launching Unity when publishing to Unity packages
- [x] Create a new exception type for handling wallet connection failures, allowing users to manage these issues
- [x] Add a user-friendly feedback editor window or menu item with a link to facilitate issue reporting for developers
- [x] Refactor the Login class to improve code organization and readability
- [x] Replace occurrences of ***Wallet with ***WalletProvider to avoid potential naming conflicts (e.g., WalletConnectWallet)
- [x] Change `web3Builder.BuildAsync()` to `web3Builder.LaunchAsync()` to better reflect its role in triggering wallet connection procedures
- [x] Confirm whether creating a web3 instance without an ISigner is still possible
- [x] Reevaluate the necessity of Unity tests (Note: May no longer be relevant)
- [x] Generate a class for a contract using its ABI
- [x] Group all menu items under a single parent (e.g., "ChainSafe/Project Settings") for organization and consistency
- [x] Remove Upload button from More examples section

Hint: If you know a good point in time where planned refactoring would make it easy to close one of the above issues, please mention them so that we can introduce those improvements gracefully.

RyRy79261:
[...]

PR #731, opened 2023-11-14
AllNft calls & technical debt

### LINKED ISSUE #719: All Nfts call error
image

All Nft call is erroring saying some value is null, will look into this. The endpoint was deprecated afaik and we need to put the new one in.

sneakzttv:
All NFT calls are now deserializing into a nice new model from the new API and displayed in the console when called. For #719
(These were working fine until this morning when the api team closed the endpoint, I guess this comes with 2.5.2 now.
I'm removing the issue from the PR, it won't be resolved for at least a few weeks as the API team is transitioning to turbosquid.

Webgl templates now only sync check if webgl is active as the editor platform when opening server settings as per #680

Added a wallet exception and made it fire when a wallet has issues connecting from the login scene as per #680


Packages/io.chainsafe.web3-unity/Runtime/Scripts/Model/AllNftModel.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace ChainSafe.Gaming.UnityPackage.Model
+{
+    public class AllNftModel
```
kantagara, line 6: Was this necessary to be a class?

sneakzttv: yes, allowing it do be deserialized into a class gives the users heaps of freedom when navigating the json responses. It's how we've always done it with the API calls, if you check the models in the runtime folder you'll see. If you'd like it done another way just let me know <3

sneakzttv: Doesn't really matter anyway as it now needs to be redone with the API endpoint being decommissioned this morning :|

juans-chainsafe (approved):
All good!

- AllNft 🕐 (didn't test it)
- WebGL templates when I click Save settings ✅ 
- Login exception ✅  (I couldn't make it appear but login works fine)

robGG1997 (approved):

Packages/io.chainsafe.web3-unity/Editor/ServerSettings.cs
```diff
@@ -204,4 +208,25 @@ private class ValidateProjectIDResponse
...
+            WebGLTemplateSyncStatus.DoesntExist =>
+                EditorUtility.DisplayDialog("web3.unity", "Do you wish to install the web3.unity WebGL templates into your project?", "Yes", "No"),
+            WebGLTemplateSyncStatus.OutOfDate =>
+                EditorUtility.DisplayDialog("web3.unity", "The web3.unity WebGL templates in your project are out of date, would you like to update now?", "Yes", "No"),
+            _ => false,
+        };
+        
+        if (performSync)
+        {
+            WebGLTemplateSync.Syncronize();
```
RyRy79261, line 228: Typo (Though I dont see that function within this PR so might be a separate bug fix

```suggestion
            WebGLTemplateSync.Synchronize();
```

RyRy79261 (approved):
Only the mention of the Typo but otherwise looks good

PR #738, opened 2023-11-16
Template fix & Cleanup

sneakzttv:
Don't like how the webgl templates are updated, changed it to fire on load if platform is webgl as opposed to server settings.
Updating some variable types to big integers.
Cleaning up some code.

------------To Test---------------
Create a fresh project for the SDK so you're prompted to save your server settings.
Swap platforms to webgl.
Check that you are prompted for webgl templates.
Check that the server settings are repopulated into the network.js file when updating.


juans-chainsafe (commented):
My only concern here is that network.js is not always updated, for example in this situation:

1. Be on a platform that is NOT WebGL
2. Set and save Server settings
3. Switch to WebGL and download the templates
4. Check the network.js file
In this case, network.js is not updated with the server settings values

sneakzttv:
> My only concern here is that network.js is not always updated, for example in this situation:
> 
> 1. Be on a platform that is NOT WebGL
> 2. Set and save Server settings
> 3. Switch to WebGL and download the templates
> 4. Check the network.js file
>    In this case, network.js is not updated with the server settings values

Fixed, made it save config into the js file again after updating.

juans-chainsafe (approved):
Looks good! issues were fixed, I don't see any errors

robGG1997 (approved):

RyRy79261 (approved):

PR #752, opened 2023-11-21
Removed plastic using directive

sneakzttv:
Removed plastic using directive as it was unnecessary and causing compilation errors.


RyRy79261 (approved):

juans-chainsafe (approved):
Builds correctly in the editor and in WebGL too.

I also tried some samples using the Web3Auth login

kantagara (approved):

PR #786, opened 2023-12-04
Meta Mask Integration

### LINKED ISSUE #517: Drop WebGl template and interact with browser wallets directly
**Summary**
Currently, https://github.com/ChainSafe/web3.unity-webgl-template is severely limiting the functionality of SDK and increasing the size of games unnecessarily. 

Implement interaction with window.ethereum directly. Example: https://github.com/Nethereum/Nethereum/blob/master/src/Nethereum.Unity.Metamask/NethereumMetamask.jslib

This means we will lose the WalletConnect option which we used through the web interface and instead, we should implement natively in C# using existing libraries (https://github.com/WalletConnect/WalletConnectSharp) which would enable wallet to connect across all platforms and not just on the web.

**Motivation**
[Why is this feature of general interest?]
- faster game loading
- more options for game developers
- less maintenance on our side
- better error control and handling

**Blockers**
(OPTIONAL) [Do you see any blocker that needs to be done before implementing this feature?]

**Additional context**
(OPTIONAL) [Add any other context or screenshots/videos about the feature request here.]

robGG1997:
closes #517 

Note : this only works on WebGL build (not in editor)
Before Building to WebGL

- To run add `SampleLogin - MetaMask` scene to build settings at index of 0 and `SampleMain` at index 1

- Reimport WebGL templates

OR

Once Build has finished go into the build directory and edit the `index.htm` file to include this line right [here](https://github.com/Nethereum/Unity3dSampleTemplate/blob/master/webgl/index.html#L107)

- Now `Build And Run` to WebGL or just Build then you can then run the WebGL build from CMD using `"C:/Program Files/Unity/Hub/Editor/2021.3.22f1/Editor/Data/MonoBleedingEdge/bin/mono.exe" "C:/Program Files/Unity/Hub/Editor/2021.3.22f1/Editor/Data/PlaybackEngines/WebGLSupport/BuildTools/SimpleWebServer.exe" "[Build Directory]" 60606` - DO make sure that paths are correct and replace `[Build Directory]` with your own build directory

Now open http://localhost:60606/ on your browser (preferably chromium based like chrome/brave)

Once this PR is approved and ready to merge I'll add the new dependency DLLs to be copied in publish dependencies CI so they can be automatically generated.


robGG1997:
> There is basically the same issue, that we have with the WalletConnect wallet implementation - connection doesn't occur on `web3builder.LaunchAsync()` as it should. I can refactor the WalletConnect wallet implementation and you can then take it as a reference point if you want to.

Connection is actually awaited in `LaunchAsync` in Wallet Connect, it's also awaited here

juans-chainsafe (approved):
MetaMask template working awesome!

Just 1 minor thing, transactions in mint or transfer (erc20, erc721 or erc1155) is not displaying the message in the console that the transfer or mint was executed succesfully, not sure why.
The last message that I see in the console is `[Web3] Successful eth_getTransactionByHash JsonRPC response with result ChainSafe.Gaming.Evm.Transactions.TransactionResponse`

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
```diff
@@ -0,0 +1,197 @@
...
+using Nethereum.Unity.Metamask;
+using Nethereum.Unity.Rpc;
+using UnityEngine;
+
+namespace ChainSafe.Gaming.MetaMask.Unity
+{
+    /// <summary>
+    /// Controller script for connecting to MetaMask and making awaitable JsonRPC requests.
+    /// </summary>
+    public class MetaMaskController : MonoBehaviour
```
oleksandrchainsafe, line 16: Can we make it a simple plain C# object instead of MonoBehaviour?

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
```diff
@@ -0,0 +1,197 @@
...
+        private void InvokeAccountConnected(string address)
+        {
+            OnAccountConnected?.Invoke(address);
+        }
+
+        /// <summary>
+        /// Callback when a new account is selected by user.
+        /// </summary>
+        /// <param name="address">New selected account address.</param>
+        public void NewAccountSelected(string address)
```
oleksandrchainsafe, line 175: I don't think we should allow to switch accounts mid-game. At least right now there is no event nor nothing for the user to handle this.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
```diff
@@ -0,0 +1,197 @@
...
+            logger.Log($"New Account with address {address} Selected.");
+
+            ConnectedAddress = address;
+        }
+
+        /// <summary>
+        /// Callback when chain is changed on Network.
+        /// </summary>
+        /// <param name="chainId">New chain.</param>
+        public void ChainChanged(string chainId)
```
oleksandrchainsafe, line 186: We can't allow chain changing also.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+        public async Task<T> Request<T>(string method, params object[] parameters)
+        {
+            return await metaMaskController.Request<T>(method, parameters);
+        }
+
+        /// <summary>
+        /// Disconnect from MetaMask.
+        /// </summary>
+        /// <returns>Awaitable disconnect task.</returns>
+        public Task Disconnect()
```
oleksandrchainsafe, line 84: If there's no disconnection functionality available, then we can remove this method.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Implementation of <see cref="ILifecycleParticipant.WillStartAsync"/>.
+        /// Lifetime event method, called during initialization.
+        /// </summary>
+        /// <returns>async awaitable task.</returns>
+        public async ValueTask WillStartAsync()
+        {
+            Address = await metaMaskProvider.Connect();
+
+            logWriter.Log($"Connected to MetaMask account {Address}");
```
oleksandrchainsafe, line 46: Don't forget to remove/comment-out those logs before shipping.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
```diff
@@ -0,0 +1,94 @@
...
+        public MetaMaskSigner(IMetaMaskProvider metaMaskProvider, ILogWriter logWriter)
+        {
+            this.metaMaskProvider = metaMaskProvider;
+            this.logWriter = logWriter;
+        }
+
+        /// <summary>
+        /// Signer's public key/address.
+        /// </summary>
+        public string Address { get; private set; }
```
oleksandrchainsafe, line 35: We can keep this private as there is already a GetAddress() for ISigner.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSignerExtensions.cs
```diff
@@ -0,0 +1,27 @@
+using ChainSafe.Gaming.Evm.Signers;
+using ChainSafe.Gaming.Web3.Build;
+using ChainSafe.Gaming.Web3.Core;
+
+namespace ChainSafe.Gaming.MetaMask.Unity
+{
+    /// <summary>
+    /// <see cref="MetaMaskSigner"/> extension methods.
+    /// </summary>
+    public static class MetaMaskSignerExtensions
```
oleksandrchainsafe, line 10: Personally, I would keep all extension methods related to MetaMask Provider/Signer/TxExec in one class, so that there are fewer files in the project.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskTransactionExecutor.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            return await metaMaskProvider.Request<TransactionResponse>("eth_getTransactionByHash", hash);
+        }
+
+        /// <summary>
+        /// Implementation of <see cref="ILifecycleParticipant.WillStartAsync"/>.
+        /// Lifetime event method, called during initialization.
+        /// </summary>
+        /// <returns>async awaitable task.</returns>
+        public ValueTask WillStartAsync() => new ValueTask(Task.CompletedTask);
```
oleksandrchainsafe, line 85: You can remove ILifecycleParticipant from the list of parents, as there is no actual function to it.

oleksandrchainsafe (changes requested):

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskTransactionExecutor.cs
oleksandrchainsafe: Also, in other components too.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
robGG1997: Nope, the plugin needs `gameObject` name to interact with the script, It uses reflection

robGG1997: I also tried passing it as an argument but it doesn't work if the Callbacks are on a different script than what's on the `gameObject`

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
robGG1997: Metamask allows i and we can't inhibit Metamasks function, what we can do it maybe throw an exception or log an error when a new account is selected

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
robGG1997: So Metamask allows it and we can't inhibit Metamasks function, what we can do it maybe throw an exception or log an error when chain is changed

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskProvider.cs
robGG1997: The reason I kept it is because I would like Wallet Connect and Metamask providers inherit from one single class since they're both JsonRPC providers and have very similar functions

robGG1997: we can provide one single point of abstraction for all wallet providers, since all they do is connect, disconnect and request Json RPC requests

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
robGG1997: I think this is a necessary log for Devs to know, or are we logging a successful connection twice?

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSignerExtensions.cs
robGG1997: I think that's a good idea but that's not how other projects in the solution do it.

robGG1997: Should I just change this for Metamask?

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskTransactionExecutor.cs
robGG1997: yeah you're right, I thought about that too but don't you think it could come in handy in the future?

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskProvider.cs
robGG1997: should be an issue on its own but this was kind of a prelude for it but I'll remove it if you sill think I should

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
robGG1997: yep, will do

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
robGG1997: Yep we're logging twice, will remove

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
kantagara: Yeah, that should be the way to go :)

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
kantagara: Yup, do that 😁

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
robGG1997: done

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskController.cs
robGG1997: done

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
robGG1997: done

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
robGG1997: removed

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSignerExtensions.cs
kantagara: I think it should be kept like this. 
If we want to unify them (LIke walletconnectExtensions for both rovider/signer/txexec go in a single class), we should create an issue for it. But it's a low-prio imo.

kantagara (approved):

Packages/io.chainsafe.web3-unity/Editor/WebGLTemplates/Web3GL-MetaMask/index.html
```diff
@@ -103,6 +103,7 @@
         createUnityInstance(canvas, config, (progress) => {
           progressBarFull.style.width = 100 * progress + "%";
         }).then((unityInstance) => {
+          nethereumUnityInstance = unityInstance;
```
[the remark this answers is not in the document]
robGG1997: we don't really need to since it's Js, it'll declare it implicitly

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSignerExtensions.cs
robGG1997: agreed

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskProvider.cs
oleksandrchainsafe: We already have Signer and TxExec abstractions that are essentially a breakdown of Wallet functionality. Some "Wallets" provide only Signer functionality, some provide both, some can also provide RPC Provider functionality. But I don't think we need more abstractions there. Let's keep it simple.
We can, however, try to transform those custom providers into a standardized IRpcClient, that's true. Not sure if it's possible though. This needs more research.
For now, I would remove this method just to keep the codebase clean.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
oleksandrchainsafe: I mean all of the logs, not just this one. I would only keep the most important one or two like "Wallet connected", "Transaction Sent" and "Message Signed", not to spam in user's console.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSignerExtensions.cs
oleksandrchainsafe: @kantagara But that's how it's done in every other package besides WalletConnect.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskProvider.cs
robGG1997: understood, done!

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSigner.cs
robGG1997: Understood, I've removed logs I believe to unnecessary/redundant.

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskSignerExtensions.cs
robGG1997: I see, I've merged ext methods in both Metamask and WalletConnect

src/ChainSafe.Gaming.MetaMask.Unity/MetaMaskTransactionExecutor.cs
oleksandrchainsafe: We can always add it when we need it.

The rule of thumb here is to keep only the stuff that is currently being used. Again we can add what will come in handy in the future, in the future. But this way we'll keep the project clean from the stuff that never came in handy eventually.

robGG1997: alright, will do

oleksandrchainsafe (approved):

Packages/io.chainsafe.web3-unity/Editor/WebGLTemplates/Web3GL-MetaMask/index.html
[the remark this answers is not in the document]
kantagara: One more reason I despise JS 🤣

kantagara (approved):

sneakzttv:
checking again with new changes, one moment.

sneakzttv:
Good to go!

image
image

PR #793, opened 2023-12-06
Fixing editor login

sneakzttv:
Currently if the login scene isn't the first scene the SDK refuses to create a web3 instance. Removing the conditional region fixes this, it also allows the login scene selection variable to function correctly from anywhere.

--------------To Test--------------------------
Set the login scene as the second build index and try to login to confirm that it creates a web3 instance and allows scene transition.


kantagara (approved):

robGG1997 (approved):
this was actually not an ideal way to check if scene isn't login scene (checking build index) since we've multiple login scenes now, so yeah we should remove it, we're also probably going to implement a single scene system anyway, LGTM!

juans-chainsafe (approved):
Working in the editor and WebGL!

PR #806, opened 2023-12-14
Ez editor chain switching

sneakzttv:
We now have the ability to easily switch chains in the editor via drop down menus in the server settings area! Woohoo!
The values are populated dynamically from Chainlist's public github so we're always going to be up to date with new chains.

image

------------------------To Test---------------------------
Boot up the SDK as usual and go to the server settings area.
Ensure the search functionality functions as intended, you should only see results relevant to the current text input.
Select a chain plus an RPC from the drop down menus and press save.
Ensure the network.js files are updating with this information.
Test that server information is saving between window and editor restarts.
Test that some sample calls are working for standard & webgl builds (I didn't change anything here, just a quick sanity check)

Expected behaviors:
Adding a custom RPC will override the drop down menus selected RPC.
When a custom RPC set and you change chains or RPCs from the menus, it should remove the custom RPC value automatically.
Network, chainID & symbol values should all be updating dynamically as you change chains.


Packages/io.chainsafe.web3-unity/Editor/ChainInfoModel.cs
```diff
@@ -0,0 +1,63 @@
+using System.Collections.Generic;
+
+namespace ChainSafe.Gaming.UnityPackage.Model
```
robGG1997, line 3: 3 suggestions for this class
1. some of these can be structs except the main class
2. if we're deserializing the response we can just have the fields we need (chain name and id)
3. Model in namespace should be replaced with editor or move script into Runtime/Scripts/Model like all other models

Packages/io.chainsafe.web3-unity/Editor/ServerSettings.cs
```diff
@@ -1,46 +1,119 @@
...
     private string symbol;
     private string rpc;
+    private string newRpc;
 
     Texture2D logo = null;
+    
+    // Search window
+    private StringListSearchProvider searchProvider;
+    private ISearchWindowProvider _searchWindowProviderImplementation;
+    public static UnityEvent onDropDownChange;
```
robGG1997, line 46: can we reference this event from the search provider?

robGG1997 (changes requested):

Packages/io.chainsafe.web3-unity/Editor/ServerSettings.cs
sneakzttv: done

Packages/io.chainsafe.web3-unity/Editor/ChainInfoModel.cs
sneakzttv: done

Packages/io.chainsafe.web3-unity/Editor/StringListSearchProvider.cs
```diff
@@ -0,0 +1,111 @@
...
+        onSetIndexCallback?.Invoke((string)SearchTreeEntry.userData);
+        ChainSafeServerSettings instance = EditorWindow.GetWindow<ChainSafeServerSettings>();
+        if (instance != null)
+        {
+            instance.UpdateServerMenuInfo();
+        }
+        else
+        {
+            // Throw an exception when instance is null
+            throw new NullReferenceException("ChainSafeServerSettings instance is null. Cannot update server menu info.");
```
robGG1997, line 105: if you're throwing a null ref exception you can just `instance.UpdateServerMenuInfo();` without checking if instance is null

sneakzttv: done

kantagara (approved):

robGG1997 (approved):

juans-chainsafe (approved):
Looks good! just a minor issue that I will create a ticket, but is not a blocker.

PR #822, opened 2024-01-10
Custom networks for Multicall empty instead of null

### LINKED ISSUE #821: Multicall breaks if you're on an unsupported netowrk
**Describe the bug**
When you're on a chain id that's not listed on MultiCallDefaults.DeployedNetworks you get an error due to the fact that config.CustomNetworks is null

** Steps to reproduce ** 
Set an unsupported Chain ID 

**Expected behavior**
You shouldn't get an error in regards to multicall, but rather make custom networks not null and extendable (as a Dictionary or something like that).

RyRy79261:
closes #821 

For making the custom network extendable, presently, the networks are only consumed when multicall's service is initialized


kantagara (approved):

kantagara (approved):

kantagara (changes requested):
You removed the line 36 as well, but that line should've been there.

kantagara (approved):

oleksandrchainsafe (approved):

robGG1997 (approved):

sneakzttv (approved):

Issue #847, opened 2024-01-29
Ramp not working with Metamask scene login


juans-chainsafe:
**Describe the bug**
Ramp is working in all platforms, except when you use the scene **SampleLogin - Metamask**

**Steps**
1. Use the scene [SampleLogin - Metamask]
2. Build with WebGL
3. Login with [Metamask]
4. Click on "Try On Ramp"

**Expected behavior**
Ramp modal appears

**Console errors**
```
webgl.framework.js.br:3 ArgumentNullException: Value cannot be null.
Parameter name: format
  at System.String.FormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x00000] in <00000000000000000000000000000000>:0 
--- End of stack trace from previous location where exception was thrown ---
```

**Screenshots/Video**
<img width="1781" alt="Screenshot 2024-01-29 at 16 56 23" src="https://github.com/ChainSafe/web3.unity/assets/104365141/be4d9281-c252-4435-8963-dc9219cd48d6">

**Desktop (please complete the following information):**
 - Chainsafe SDK version: [main]
 - Unity Version: [e.g., 2022.1]
 - Operating System: [Mac]

kantagara:
@juans-chainsafe this is a regression that happened after Robs push to main. Will fix.

PR #855, opened 2024-02-06
Fixing Ramp Not Working with the new scene setup

### LINKED ISSUE #844: Remove traces of Ramp from the main package
**Summary**
Currently, it's a bit ugly how ramp is working and I am fully aware of it-> it is a spearate module but I still referenced it in the main scene just for the sake of making everything work across all platforms on which we support Ramp.

Rob is going to create a nice adapter pattern that we are going to utilize for injecting additional services per our (and our developers) needs.

**Motivation**
Cross-referencing modules especially those that are not shipped together is not a good idea.

**Acceptance Criteria**
- [] Have RampServiceProvider be easily drag'n' dropped in the login scene

**Blockers**
- [] Lack of the actual implementation in the main branch.

**Dependencies**
- [] This issue can be worked on only when we can actually have the interface/scriptable object on the main branch.

kantagara:
This closes #844 and #847 

Will also update documentation and record a new video.


rob1997:
Are we just adding the ramp service adapter? we could add #if platform dependent compilation for it too since ramp currently only works on specific platforms

kantagara:
Oh good call. I will also add two more scripts!

kantagara:
@rob1997 added it.

rob1997 (approved):

rob1997:
> @rob1997 added it.

LGTM! what's the use case for the new script? disabling GO based on available service

kantagara:
> > @rob1997 added it.
> 
> LGTM! what's the use case for the new script? disabling GO based on available service

If you log in on main, you will see Ramp being there (in Editor), which is not supported platform. That scripts ensures that game objects that are bound to certain services will not show up in the SampleMain because that would give you errors..

sneakzttv (approved):
One step closer to 2.6 woo, love how you're tieing a nullcheck boolean on services into activation btw, top tier.

```
private void Awake()
    {
        gameObject.SetActive(Web3Accessor.Web3.ServiceProvider.GetService(_typesDictionary[serviceType]) != null);
    }
```

juans-chainsafe (approved):
Ramp working correctly!
Metamask scene ✅ 
iOS ✅ 
WebGL ✅

PR #856, opened 2024-02-06
Analytics now send correct events

### LINKED ISSUE #854: Analytics Bug - not sending right RPCs
**Describe the bug**
It seems that, even tho I replaced just the analytics client, that names of the methods are not going trough properly. 

**Steps**
1. Send balanceOf on ERC20
2. Expect balanceOf to be sent out
3. You'll see just chain_id being sent out
**Expected behavior**
you'd see balanceOf in countly analytics

 - Chainsafe SDK version:main
 - Unity Version: 2021
 - Operating System: any

kantagara:
This closes #854 

I really couldn't make the ContractBuilder send analytics events because it gets assigned as soon as Web3Builder gets created (I guess it's necessary), and that throws some exceptions. So I had to rework it in a way that I go trough the concrete implementation of the IRpcProvider which I know has the reference to the analyitics client.


sneakzttv (approved):
Woo shell file via editor

image

Nice one parsing the provider too
image

juans-chainsafe (approved):
This looks great now! There are a few things that doesnt have any logs at all:

- Sha3
- SignMessage
- SignVerify
- EcdsaSignMessage
- EcdsaSignTransaction
- EcdsaGetAddress

In case we can add log for those, I can retest it or have a different ticket

Packages/io.chainsafe.web3-unity/Runtime/Scripts/CountlyAnalytics.cs
```diff
@@ -14,6 +14,7 @@ public class CountlyAnalytics : IAnalyticsClient
     
     public async void CaptureEvent(AnalyticsEvent eventData)
     {
+        Debug.Log($"Capturing event: {eventData.EventName}");
```
oleksandrchainsafe, line 17: We should probably hide this from the user (read 'remove').

src/ChainSafe.Gaming/Web3/Evm/JsonRpc/RpcClientProvider.cs
```diff
@@ -17,26 +17,38 @@ public class RpcClientProvider : IRpcProvider, ILifecycleParticipant
...
+            this.analyticsClient = analyticsClient;
+            this.projectConfig = projectConfig;
 
             if (string.IsNullOrEmpty(this.config.RpcNodeUrl))
             {
                 this.config.RpcNodeUrl = chainConfig.Rpc;
             }
         }
 
+        public IAnalyticsClient AnalyticsClient => analyticsClient;
```
oleksandrchainsafe, line 46: To get rid of those properties you can add one more required parameter to the list of ContractBuilder's dependencies. This basically means adding the IAnalyticsClient parameter to every constructor overload. You can then store it and pass it to the Contract's constructor in Build.
IAnalyticsClient should be always registered in DI by design, that's why you don't need to create a new overload for ContractBuilder and can simply add this to every existing constructor overload.

oleksandrchainsafe: To access IProjectConfig and IChainConfig you can reference them as dependencies in the implementation of your IAnalyticsClient.

oleksandrchainsafe (dismissed):

src/ChainSafe.Gaming/Web3/Evm/JsonRpc/RpcClientProvider.cs
kantagara: I Swear I tried to do that exact thing but for whatever reason my project was constatnly failing :D 
now I asked GPT to do that for me and it works like a charm.....

rob1997 (approved):

PR #868, opened 2024-02-13
Sygma Client Types

kalambet:
This is the first PR to support Sygma in the Gaming SDK.

This PR **only contains Types and no logic** at all.
Some of the classes might be transformed into structures as logic gets implemented.

With the next PRs I'll be adding each of the Sygma Client functions implementations so it is not too overwhelming.


rob1997 (approved):
LGTM!

sneakzttv (approved):

src/ChainSafe.Gaming.SygmaClient/SygmaClient.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.contractBuilder = contractBuilder;
+            this.signer = signer;
+            this.sourceChainConfig = sourceChainConfig;
+            this.destinationChainConfig = destinationChainConfig;
+            this.analyticsClient = analyticsClient;
+            this.projectConfig = projectConfig;
+        }
+
+        public bool Initialize()
+        {
```
kantagara, line 36: Is this necessary? If it's an ilifecycle participant we have WillStartAsync

kantagara (approved):

src/ChainSafe.Gaming.SygmaClient/SygmaClient.cs
kalambet: But you are right, method arguments are missing. Let me update it.

src/ChainSafe.Gaming.SygmaClient/SygmaClient.cs
```diff
@@ -0,0 +1,79 @@
...
+            throw new System.NotImplementedException();
+        }
+
+        public Task<TransferStatus> TransferStatusData(Environment environment, string transactionHash)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public ValueTask WillStartAsync()
```
oleksandrchainsafe, line 69: Usually, we keep those at the top of the methods list as part of the initialization/termination method group.

src/ChainSafe.Gaming.SygmaClient/Types/ConfigUrl.cs
```diff
@@ -0,0 +1,21 @@
+namespace ChainSafe.Gaming.SygmaClient.Types
+{
+    public static class ConfigUrl
+    {
+        public static readonly string Devnet = "https://chainbridge-assets-stage.s3.us-east-2.amazonaws.com/shared-config-dev.json";
```
oleksandrchainsafe, line 5: You can use `const` instead of `static readonly` for strings.

oleksandrchainsafe (approved):
👍

src/ChainSafe.Gaming.SygmaClient/SygmaClient.cs
kalambet: Will change it with the next PR! Thanks!

src/ChainSafe.Gaming.SygmaClient/Types/ConfigUrl.cs
kalambet: Will do, thank you!

PR #869, opened 2024-02-14
Generic types for arrays

sneakzttv:
Altered the getarray method to accept arguments if they're not null & altered the data type from string[] to a generic type so that all data can be parsed. This was giving me issues with Blasters and racers as my data types aren't string arrays. All we need to do now is explicitly declare the datatype when making the getarray call which you can see in the test script.

Solidity isn't meant to deal with strings technically which makes it sub optimal for gas and should be avoided where possible.

I can pull data correctly with these changes as you can see the type below as biginteger & it's processing fine.
image

To Test:

Check the sample scene and ensure both get and send array functions still work (they do)


juans-chainsafe (approved):
Working with metamask and web3auth !

kantagara (approved):

PR #876, opened 2024-02-27
Sygma iteration 2 - Made the BuildApprovals method. Added method in contract that helps me build a transaction request

kantagara:


src/ChainSafe.Gaming/RPC/Contracts/Contract.cs
```diff
@@ -161,20 +161,8 @@ public async Task<object[]> Send(string method, object[] parameters = null, Tran
             parameters ??= Array.Empty<object>();
 
             var function = contractBuilder.GetFunctionBuilder(method);
-            var txReq = overwrite ?? new TransactionRequest();
```
kalambet, line 164: I think this change breaks the original logic. The optional assignment for `From`, `To`, `Data` and etc was there in the case provided `overwrite` object does not have any.

So it was possible to provide some user-defined fields in this method. But with this logic `overwrite` object is not used at all.

src/ChainSafe.Gaming/RPC/Contracts/Contract.cs
```diff
@@ -88,7 +88,7 @@ public async Task<object[]> Call(string method, object[] parameters = null, Tran
             var function = contractBuilder.GetFunctionBuilder(method);
             var txReq = overwrite ?? new TransactionRequest();
             txReq.To ??= address;
-            txReq.From ??= await signer.GetAddress();
```
kalambet, line 91: Maybe adopt `PrepareTransactionRequest` here as well?

src/ChainSafe.Gaming/RPC/Contracts/Contract.cs
kantagara: You're right! Fixed

rob1997 (approved):
some DRY in `PrepareTransactionRequest` method, LGTM!

sneakzttv (approved):

kalambet (approved):
Nice!

kalambet (approved):

PR #886, opened 2024-03-01
Fixed #885 - DefaultValueHandling for TranscactionRequest

### LINKED ISSUE #885: eth_call: INVALID_PARAMS | eth_call Tx is serialized with null fields, RPC server returns error on chains such as any SKALE hub
### Description

When making a call to a contract on a chain such as a SKALE hub (e.g. SKALE Nebula Hub Testnet) using ``Contract.Call(string method, object[] parameters, TransactionRequest overwrite)``

A TransactionRequest object is created and its variables are set appropriately as given. That object is passed on eventually to the HttpClient to be serialized and sent to the RPC provider. 

TransactionRequest is serialized with Newtonsoft, and when it is serialized with something like ``JsonConvert.SerializeObject`` as seen on the UnityHttpClient implementation, any missing fields are set to null, giving a JSON string that can look like this:

```json
[
    {
        "from": "<signer_address>",
        "to": "<contract_Address>",
        "gas": null,
        "gasPrice": null,
        "maxFeePerGas": null,
        "maxPriorityFeePerGas": null,
        "value": null,
        "data": "<function_data_encoded>",
        "nonce": null,
        "accessList": null
    },
    "latest"
]
```

RPC servers receive this and return ``Error for "eth_call": -32602 INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised``

### Expected Behavior

The parameters should be converted to JSON without including any nulls for fields which were not set. RPC server would recognize it as a proper valid request and send the correct response. 

The correct expected JSON for the above example should be:

```json
[
    {
        "from": "<signer_address>",
        "to": "<contract_Address>",
        "data": "<function_data_encoded>",
    },
    "latest"
]
```

### To Reproduce

Switch network to SKALE Nebula Hub Testnet, make any "eth_call" request to any deployed contract.

fourthedesign:
Added ``DefaultValueHandling = DefaultValueHandling.Ignore`` to ``TransactionRequest``'s ``JSONProperty`` attributes.
 
 This makes TransactionRequest ignore empty fields when serialized, and solves issue #885


fourthedesign:
https://github.com/ChainSafe/web3.unity/pull/886/commits/e06e9256356695330eeb4e884fa7fef03fe5c95f 

Adds a new method on ``Contract.PrepareTransactionRequestNoGasInfo`` that does not get and add gas info to the transaction for the eth_call. This both optimizes the call and also solves a subsequent null error where chains like SKALE do not contain fields like baseFeePerGas.

fourthedesign:
> [e06e925](https://github.com/ChainSafe/web3.unity/commit/e06e9256356695330eeb4e884fa7fef03fe5c95f)
> 
> Adds a new method on `Contract.PrepareTransactionRequestNoGasInfo` that does not get and add gas info to the transaction for the eth_call. This both optimizes the call and also solves a subsequent null error where chains like SKALE do not contain fields like baseFeePerGas.

Issue for the null error referenced here:
#887

juans-chainsafe (approved):
Tested all samples in Web3Auth and Metamask, both in WebGL with Sepolia, didn't find issues.

src/ChainSafe.Gaming/RPC/Contracts/Contract.cs
```diff
@@ -85,7 +85,7 @@ public async Task<object[]> Call(string method, object[] parameters = null, Tran
 
             parameters ??= Array.Empty<object>();
 
-            var txReq = await PrepareTransactionRequest(method, parameters, overwrite);
+            var txReq = await PrepareTransactionRequestNoGasInfo(method, parameters, overwrite);
```
kantagara, line 88: But now you would eliminate all the POST London forks. What I would suggest is to actually do something like this: if providers GetFeeData returns a null, or a specific object inside of it is null, you just return the PrepareTransactionRequestNoGasInfo. That way we're utilizing both functions.

sneakzttv (approved):

sneakzttv:
reopened here due to access rights https://github.com/ChainSafe/web3.unity/pull/897

PR #896, opened 2024-03-08
Add ERC1155 Support

### LINKED ISSUE #889: Add ERC1155 support for `SygmaClient`
**Summary**
Add ERC1155 support to the `SygmaClient`.

`CreateFungibleTransfer` methods need to be added to the `SygmaClient` but ERC1155 specifics also need to be taken into account. It is not only the _token ID_ but also the _amount_ that needs to be passed.

Support for `ERC1155Handler` and `ERC1155Safe` also should be implemented.
The fee handler should remain `Basic`.

**Motivation**
All the contracts that got deployed are not ERC721 but ERC1155 so the `SygmaClient` functionality should be accommodated.

**Acceptance Criteria**
All the basic logic for the 1155 support needs to be added. So far no testing is needed. There will be a separate issue for that.

kalambet:
This PR fixes #889


sneakzttv:
abi conversion is failing you have a new line operator in the 2 abis, maybe that? 
image
image

kantagara (approved):

kalambet:
> abi conversion is failing you have a new line escape sequence in 2 of the abis, maybe that?

Thank you for noticing this Sneakz. For the ABI it would not matter, it is an escaped sequence and will just be ignored. The reason was that some older libraries were temporarily used for testing.

sneakzttv:
> > abi conversion is failing you have a new line escape sequence in 2 of the abis, maybe that?
> 
> Thank you for noticing this Sneakz. For the ABI it would not matter, it is an escaped sequence and will just be ignored. The reason was that some older libraries were temporarily used for testing.

yeah i was curious if it would see it as an esc sequence or a raw value in an abi string, i updated it anyway and im guessing the libs updated too with nikolas update from main. Just a coincidence that it worked then.

sneakzttv (approved):

rob1997 (approved):

PR #897, opened 2024-03-08
DefaultValueHandling for TranscactionRequest

### LINKED ISSUE #885: eth_call: INVALID_PARAMS | eth_call Tx is serialized with null fields, RPC server returns error on chains such as any SKALE hub

sneakzttv:
Copy of https://github.com/ChainSafe/web3.unity/pull/886 with a slight tweak to handle both cases
Closes #885

Thanks to fourthedesign


boorich (approved):
🙏🏼

rob1997 (approved):

PR #904, opened 2024-03-13
Web3auth seperate window for webgl

### LINKED ISSUE #894: Web3Auth -> Ensure it's opening in the new tab in webgl.
**Summary**
This issue is critical for any game that builds using our SDK and is not immediately on login scene like our sample is.

**Motivation**
Testing out block racers on webGL gives us that problem and the best solution would be to do web3auth in the new tab instead of opening it in the same tab.

**Acceptance Criteria**

- [ ] Web3auth tab opens
- [ ] User logs in 
- [ ] Our game logs in successfully.

sneakzttv:
Web3auth now opens in a seperate window for webgl so that it doesn't close the game.

Closes #894 

To test:
Login in with web3auth in webgl.

Expected behavior:
A new window will open to let you connect.


juans-chainsafe (changes requested):
There is a build error on this PR:
```
Library/PackageCache/com.tools.webglthreadingpatcher@c2f3628a02/ThreadingPatcher/Editor/WebGLThreadingPatcher.cs(21,38): error CS1061: 'BuildReport' does not contain a definition for 'GetFiles' and no accessible extension method 'GetFiles' accepting a first argument of type 'BuildReport' could be found (are you missing a using directive or an assembly reference?)

Unloading broken assembly Library/ScriptAssemblies/Unity.Rider.Editor.dll, this assembly can cause crashes in the runtime
```

juans-chainsafe:
> There is a build error on this PR:
> 
> ```
> Library/PackageCache/com.tools.webglthreadingpatcher@c2f3628a02/ThreadingPatcher/Editor/WebGLThreadingPatcher.cs(21,38): error CS1061: 'BuildReport' does not contain a definition for 'GetFiles' and no accessible extension method 'GetFiles' accepting a first argument of type 'BuildReport' could be found (are you missing a using directive or an assembly reference?)
> 
> Unloading broken assembly Library/ScriptAssemblies/Unity.Rider.Editor.dll, this assembly can cause crashes in the runtime
> ```

Sorry, my bad, it was on my editor, solved uninstalling WebGLThreadingPatcher package and install it again

juans-chainsafe (commented):
@sneakzttv @kantagara Is working, so is opened in a new tab but after the user login in this new tab, the first tab remains there with the web3auth login scene, and the user is logged in the second tab, is the expected behaviour? Basically, which one should happen:

- User is logged in the second tab (and the first tab remains there with the login scene)
- Second tab is closed after the user login and first tab redirects to the SampleMain
- User is logged in the second tab and the first tab is closed automatically to not cause confusion to the user

kantagara:
> Second tab is closed after the user login and first tab redirects to the SampleMain

This should be the expected behaviour.

sneakzttv:
i don't think we have access to the auth page, i think the redirect back to the url is happening there, it's also changing from localhost to localhost + session id

sneakzttv:
image
image

PR #917, opened 2024-03-26
Added Generic<T> Extension Methods for Contract

oleksandrchainsafe:
Original Call & Send methods of Contract class operate with arrays of object `object[]`, for both input and output values. This leads to a lot of boilerplate code each time you want to interact with a smart-contract.

The extension methods automate most of this work for you.


kantagara (approved):
What if there's more than 3 params that are being passed on? :D

oleksandrchainsafe:
> What if there's more than 3 params that are being passed on? :D

Then there are 2 options:
- Do it using the old ugly way,
- Create your own extension method with 4, 5, 6.. parameters (and ideally post it as PR)

But from my experience so far, 3 arguments are more than enough for 99% of the cases.

src/ChainSafe.Gaming/RPC/Contracts/ContractExtensions.cs
rob1997, line 1: Might I suggest a more robust solution using `params` in `Contract.Call` and `Contract.Send` Methods such as this
`public async Task<TOut> Call<TOut>(string method, TransactionRequest overwrite = null, params object[] args)` `TOut` is the return type, we don't need a specific type for lists because we can just replace `TOut` with `IEnumerable<>`

sneakzttv (approved):

src/ChainSafe.Gaming/RPC/Contracts/ContractExtensions.cs
oleksandrchainsafe: We can add the overload with dynamic count and types as input, I guess.

oleksandrchainsafe: We'd still need a 'Call/SendMany' methods due to the specifics of how smart contracts typically return data.
When a smart contract method returns a list of values, it returns a JSON list of values.
When a smart contract returns one value, it still returns a list of values, but with just one value in it.
We automatically fetch the first value from the array in this case, as you can see in the code. If you try to pass IEnumerable as the type argument, the method will try to cast the first element of the response array to the IEnumerable. This is not what one would expect from the method. Therefore there are two different methods for each case + the void overload for Send.

rob1997: hmm for Call/SendMoney you're right extensions methods seems like the best way to go but for the rest (multiple inputs) this'll work fine and I don't mean to add an overload/another method, we can replace the current call/send with this. The params will allow you to call/send with as many different inputs as you want without casting types at all. 
But it's up to you, free to resolve this conversation :)

PR #935, opened 2024-04-30
Sygma directive fix

sneakzttv:
Sygma directive causing errors with IOS builds.


kantagara (approved):
Merge to main!

Packages/io.chainsafe.web3-unity/Runtime/Common/Scripts/ILoginProvider.cs
rob1997, line 1: This isn't a compile error it's a runtime error, are we sure this is the fix? I don't have an IOS device to check

rob1997 (approved):

PR #954, opened 2024-05-14
Ipfs upload duplicate fix

### LINKED ISSUE #951: Throw error if upload file or metadata to ChainSafe IPFS fails
**Describe the bug**
If you are using IPFS samples and a failed response happens from Storage API, we are not catching that error and we return the IPFS anyways, so the dev doesn't now that something failed.

**Steps**
1. In SampleMain set in IPFS Calls script the Bucket ID and API key from https://app.storage.chainsafe.io/
2. Use the scene [SampleLogin or SampleLogin - Web3Auth]
3. Login with [Web3Auth or Remote wallet]
4. Click on "Upload Metadata" sample
5. Click again in "Upload Metadata" sample

**Expected behavior**

- An error is displayed based on the API response due that there is a conflict with the file (because is duplicated)
- We need to verify the "status" field inside "file_details" object in the API response and show the "message" field if status is "failed"

**Console errors**
This is the API response in case of a conflict:
```
{
    "path": "/test",
    "files_details": [
        {
            "path": "/test/package.json",
            "cid": "QmbqtW6GFZofCGa41rqDhhyznFLKoGZ5LPSBAhxFeGH2gV",
            "content_type": "application/json",
            "size": 2103,
            "status": "failed",
            "error_code": 409,
            "message": "path:/test/package.json, conflict, entry can't be added",
            "suggestion": "/test/package(1).json"
        }
    ]
}
```

**Screenshots/Video**
329930806-68bff7d7-b541-4aa5-bf81-807949638209

sneakzttv:
Made an error throw if the web request returns an error status due to duplicate files already existing, displaying the message to the user also. closes #951 

Webgl still needs to be done as it requires a JS function to go into the compiler conditional but i'll fix that after the marketplace. Added a workaround to fix builds for now.

To test:
Upload a file twice using the GUI in the sample scene.

Expected result:
Error is thrown on duplicate upload.


kantagara (approved):

Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/IPFS/IPFS.cs
```diff
@@ -163,6 +174,12 @@ private static async Task<string> Upload(string apiKey, string bucketId, string
             {
                 throw new WebException(requestUpload.error);
             }
+            
+            var uploadResponse = JsonConvert.DeserializeObject<Path>(requestUpload.downloadHandler.text);
+            if (uploadResponse.files_details[0].status == "failed")
+            {
+                throw new Exception($"Upload Failed: {uploadResponse.files_details[0].message}");
```
oleksandrchainsafe, line 181: We can throw Web3Exception here I think as it's specific to our SDK.

oleksandrchainsafe (approved):
👍👍👍

Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/IPFS/IPFS.cs
sneakzttv: good idea

juans-chainsafe (approved):
Got expected error if file is duplicated:
`Web3Exception: Upload Failed: path:/Logo.png, conflict, entry can't be added`

Awesome! I will create a different ticket for webgl

PR #971, opened 2024-05-23
Adding remember me to hyperplay

sneakzttv:
Adding remember me to hyperplay

### DIFF
--- a/src/ChainSafe.Gaming.HyperPlay/HyperPlayExtensions.cs
+++ b/src/ChainSafe.Gaming.HyperPlay/HyperPlayExtensions.cs
@@ -1,3 +1,4 @@
+using System.Runtime.CompilerServices;
 using ChainSafe.Gaming.Evm.Signers;
 using ChainSafe.Gaming.Web3.Build;
 using ChainSafe.Gaming.Web3.Core;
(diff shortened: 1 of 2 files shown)


src/UnitySampleProject/Assets/Resources/HyperPlayConfigData.asset
```diff
@@ -0,0 +1,17 @@
...
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: c2292ae192d34362ac76a388d530dbb4, type: 3}
+  m_Name: HyperPlayConfigData
+  m_EditorClassIdentifier: 
+  storedSessionAvailable: 0
+  storedWallet: 
+  rememberMe: 1
```
juans-chainsafe, line 17: small change, this should be 0, to have the same behaviour in all the "Remember me" checkboxes. By default the checkbox should not be selected

juans-chainsafe (changes requested):
Remember me is not working as expected because session is not restored 

I see this output when I start the login scene again, after doing a login previously:
Screenshot 2024-05-23 at 11 18 15 AM

Maybe @rob1997 you have some idea on something else that is required for Hyperplay

src/UnitySampleProject/Assets/Resources/HyperPlayConfigData.asset
sneakzttv: yeah, still testing, ignore this PR for now <3

PR #985, opened 2024-05-29
Countly - Removing redundant segmentation and tying everything to a specific user.

### LINKED ISSUE #968: Countly - Don't send Project ID with every event but rather send it with the user ID.
**Summary**
Countly - Don't send Project ID with every event but rather send it with the user ID.

**Motivation**
Analytics team asked me to do it like that 😄 

**Acceptance Criteria**
Project ID should appear with a user rather than sent via every single event

kantagara:
This closes #968


sneakzttv (approved):

juans-chainsafe (approved):
projectId is being added to user profile in Countly as expected and projectId is not being added everywhere.

But there is a issue that is not logging events, but is happening in dev as well.

rob1997 (approved):

PR #991, opened 2024-05-30
Nikola/server settings automatic save

kantagara:
Wanted to rewrite the entire thing but it was consuming a lot of time so I opted in for a much simpler approach where I'm just patching things 😄


juans-chainsafe (approved):
Working! I see data changing in real time, and errors if data is wrong.
As well I checked in Countly that the new projectId is displayed.

I did a WebGL build, just in case

sneakzttv (approved):

rob1997 (approved):

PR #1043, opened 2024-06-27
Marketplace auth system & browsing calls

sneakzttv:
Marketplace auth system now works, login with a wallet or web3auth to test it (set target login scene to SampleMarketplace)
Browsing calls are populating items.

Bugs I'm aware of that I can get to later:

Refresh token call firing instantly (minor)
Colours not carrying over when logging out (minor)


rob1997 (approved):
Just took a glance, needs to go in since it's a blocker and has a deadline.

src/UnitySampleProject/Assets/Samples/web3.unity SDK Marketplace/2.6.0/Marketplace Samples/Scripts/AuthSystemManagerMarketplace.cs
```diff
@@ -0,0 +1,62 @@
...
+using UnityEngine;
+using UnityEngine.Networking;
+using ChainSafe.Gaming.Marketplace.Models;
+
+namespace ChainSafe.Gaming.Marketplace
+{
+    /// <summary>
+    /// Auth system manager to help with refresh tokens.
+    /// </summary>
+    public static class AuthSystemManagerMarketplace
```
kantagara, line 12: Maybe exclude the manager name in here.

kantagara: Or no, why is this not a basic class that gets injected trought the DI?

sneakzttv: it has ambiguous naming with the response class so i made it that

kantagara: I mean you can just call this one MarketplaceAuth. And inject it as a dependency. Have the IMarketplaceAuth interface and do the standard thingy with our extendibility.

PR #1071, opened 2024-07-29
Nikola/wc prompt network switch

### LINKED ISSUE #1070: Switch/Assert Chain ID of Wallet
_Creating the ticket so we won't forget. Ask me @oleksandrchainsafe if you need more info._

Right now we don't switch nor check if the chain id set in the project config matches the chain id of the wallet (at least with WalletConnect->MetaMask setup). This can lead to the situation when user sends transaction to the wrong chain, if he didn't select the right chain manually (again tested in MM only).

kantagara:
Closes #1070  #1072


src/ChainSafe.Gaming.WalletConnect/WalletConnectProvider.cs
```diff
@@ -233,6 +242,64 @@ public override async Task<string> Connect()
...
+        {
+            var chainId = GetChainId();
+            if (chainId != $"{ChainModel.EvmNamespace}:{chainConfig.ChainId}")
+            {
+                await PromptNetworkSwitch();
+                UpdateSessionChainId(chainConfig.ChainId);
+            }
+        }
+
+        private void UpdateSessionChainId(string newChainId)
```
oleksandrchainsafe, line 255: `newChainId` is not used

oleksandrchainsafe: what does this method do?

src/ChainSafe.Gaming.WalletConnect/WalletConnectProvider.cs
```diff
@@ -233,6 +242,64 @@ public override async Task<string> Connect()
             }
         }
 
+        private async Task CheckAndSwitchNetwork()
+        {
+            var chainId = GetChainId();
+            if (chainId != $"{ChainModel.EvmNamespace}:{chainConfig.ChainId}")
```
oleksandrchainsafe, line 248: I think we should recursively check if the chain id is correct till it's correct here.

oleksandrchainsafe (changes requested):

src/ChainSafe.Gaming.WalletConnect/WalletConnectProvider.cs
kantagara: Well, essentially, from my extensive debugging session, I've figured that the namespaces and accounts, after switching the network, now actually have 2 networks. The first one is the network with whom you've connected with (if you were on cronos, cronos will be the first element in the arrray). With updating the sessions chain id, we're completely removing the chain with which we've connected with originally, leaving us with the correct chain.

src/ChainSafe.Gaming.WalletConnect/WalletConnectProvider.cs
kantagara: You're right.

kantagara: Actually, no 😁  we are prompting them to switch, if they reject, we don't care that much because _every_ transaction that will happen in the future, will prompt them to switch to the corect chain due to the UpdateSessionChainId function being called right after it, which essentially sets the correct chain.

oleksandrchainsafe: aha, good then

sneakzttv (approved):

oleksandrchainsafe (approved):
GJ

juans-chainsafe (commented):
This is working in the editor as expected.

The problem is that we have a issue in iOS since we change the Unity version to 2022.3.4, I wasn't able to change to 2022.3.26 and test it there, there is a different error in Unity (I guess something else needs to happen to be able to downgrade).

I created this ticket https://github.com/ChainSafe/web3.unity/issues/1072 for the iOS issue and is up to the team if we want to merge this to dev. I can retest it on iOS together with the iOS fix.

juans-chainsafe (approved):
With the switch of Unity version:

- iOS build ✅ 
- Switch network in iOS ✅ 

The behaviour of switching the network in iOS is different from the editor, but works as expected! In this case, the network is switched when you interact with a sample that requires to sign a transaction, so the switch is done first.

Awesome work!

PR #1074, opened 2024-07-30
Metamask DLL should only be available on webGL

### LINKED ISSUE #1073: Exclude metamask dll from iOS after every reimport
**Describe the bug**
Currently, after each setup.sh run, we have to manually exclude Chainsafe.Gaming.Metamask.dll from the platform so that our iOS can compile properly (if we're on a version higher than 2022.3.11) 😄 because we are deleting the .meta files for all the dlls for our Chainsafe.Gaming solution. 
Solution is to add a simple asset preprocessor to the UnitySampleProject (and NOT in the package itself), so that we can exclude (or include) platforms for certain dlls
**Steps**
1. Run setup.sh 
2. Try to build iOS on any version of Unity newer than 2022.3.10
3. Notice the compile error.

**Expected behavior**
iOS should compile normally

kantagara:
Closes #1073


sneakzttv (approved):

juans-chainsafe (approved):
webGL and iOS working in 2022.3.26 Unity version

rob1997 (approved):

PR #1088, opened 2024-08-07
Server settings not saved

kantagara:
also added a scroll bar so that it's nice.


juans-chainsafe (approved):
Server settings is saving changes now and scroll view is there.

PR #1099, opened 2024-08-13
Marketpalce prefab minor fixes

### LINKED ISSUE #1066: Minor issues with Marketplace prefab
List of things to fix

- Description shouldn't be mandatory for creating a Marketplace
- Description shouldn't be mandatory for creating a Collection
- I can't enter Marketplaces or Collections that are empty because of `Index out of bonds error` this is because is trying to get elements of marketplace/collections but fails because is empty
- In the future, will be nice to be able to add additional properties and additional files when minting a new NFT (like in the dashboard)

sneakzttv:
Closed loading menu if no items are found in collections & marketplaces.
Allowed collections & marketplaces to be created without a description to match the dashboard's flow.
Updated form data string from erc721/1155 to ERC721/1155 as the API had been altered and was causing it to throw.
Optimized & cleaned up some code for better performance.
Removed redundant using directives.
Closes #1066

To Test:
Drop the marketplace prefab into a scene
Login to the SDK and go to the scene with the marketplace prefab

Check conditions:
1. Marketplaces and collections can be created with a blank description.
2. Empty collections and marketplaces should now toggle the loading menu off when opened.

### DIFF
--- a/Packages/io.chainsafe.web3-unity.marketplace/Runtime/Scripts/Marketplace/Marketplace.cs
+++ b/Packages/io.chainsafe.web3-unity.marketplace/Runtime/Scripts/Marketplace/Marketplace.cs
@@ -167,7 +167,6 @@ namespace Scripts.EVM.Marketplace
             var formData = new List<IMultipartFormSection>
                 {
                     new MultipartFormDataSection("name", _name),
-                    new MultipartFormDataSection("description", _description),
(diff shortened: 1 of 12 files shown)


juans-chainsafe (approved):
Issues were fixed and working in the editor!

There is a CORS error in WebGL but is a different thing, I will create a separate ticket for that

Issue #1106, opened 2024-08-22
Login issues


sneakzttv:
I cant click any buttons with connect on initialize defaulted to on btw, scene does nothing till it's turned off.
https://gyazo.com/806ca2e6d935cec63e9548c558b2da28

web3auth seems to have a mind of its own, sometimes it loads, sometimes it doesn't, I have to keep going back and fourth to get the window to pop up in the browser.
https://gyazo.com/62ad1c0c2487cf60d8a0e745d2914092

takes up a decent amount of time getting through it when testing atm

PR #1109, opened 2024-08-26
Bug fixes and Improvements

rob1997:
Providers now read/write
Login issues fixed
intermittent web3auth login issue also fixed


kantagara (approved):

Packages/io.chainsafe.web3-unity/Editor/ReadWriteProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+using UnityEditor;
+using UnityEngine;
+
+namespace ChainSafe.GamingSdk.Editor
+{
+    /// <summary>
+    /// Make connection providers read-write by moving it from packages into Assets/Resources.
+    /// </summary>
+    [InitializeOnLoad]
+    public class ReadWriteProvider
```
kantagara, line 13: Don't think this makes much sense imo. Because, for people importing the package, the files would be in the package cache. Which cannot be deleted permanently.

kantagara: Why is it important to make them read/write?

rob1997: because we'll need to edit the values on the scriptable objects and if they're in packages they can't be edited, I guess I could change the name to `EditableConnectionProviders`, I think we should keep this script because when we add a new package like HyperPlay it'll automatically add the provider

kantagara: Actually, what I meant is, that we shouldn't add ConnectionProviders to the Resources folder in the package itself. But rather, let them sit in the root or in a folder named ConnectionProviderData, and transfer them from there to the resources inside of the Assets folder, because otherwise you'd get duplicate providers..

rob1997: I actually had a dilemma about this because providers are like configs and if that's the case shouldn't we make user configure/create them? I suppose we can have a button under providers in connection handler like setup/configure for each type of provider to make it easier but configuring should be for dev imo for example we need the user to setup their own wallet connect project on their dashboard and get the id to reference in the config same for web3auth so shouldn't we let users configure their own providers, let me know what you think

kantagara: Actually, yes, that is the best solution, and kind of like a standard procedure. So less hassle for us.

oleksandrchainsafe (approved):
seems legit

juans-chainsafe (approved):
I wasn't able to reproduce errors in https://github.com/ChainSafe/web3.unity/issues/1106 
Unless I don't open HyperPlay, which throws a error in the console, so Connect button cannot be clicked (but we have the remember me issue for this)

PR #1116, opened 2024-08-27
Remember me issue fix

### LINKED ISSUE #1085: Merge configs to Connection Providers
**Summary**
Currently there's more than one config per provider, we can unify those configs under connection provider which references all configs and is also a scriptable object.

Move config fields and variables into the respective connection providers.

**Motivation**
A lighter more readable and simpler SDK

**Acceptance Criteria**
There should be one Scriptable Object per provider, in case of larger (context specific) configs we can make then serializable classes/structs serialized in connection providers.

**Blockers**
(OPTIONAL) [Do you see any blocker that needs to be done before implementing this feature?]

**Dependencies**
(OPTIONAL) [Do you know about dependencies that directly affect this issue?]

**Additional Context**
(OPTIONAL) [Add any other context, files, images (screenshots/videos), diagrams, or documents about the feature request here.]

**Final Steps**
* Give it one of these labels `Type: Feature` `Type: Enhancement`
* Add your estimation

**Estimations**

| Story Point | Amount of effort | Amount of time | Task Complexity | Task Risk or Uncerntainty |
|----------|----------|----------|----------|----------|
| 3 | Mild efort | A day | Low complexity | Low |

### LINKED ISSUE #1103: Disable Remember me not working as expected
**Describe the bug**
If you do the login with remember me enabled, and then you disable the remember me but try to login again, it will remember the previous session.

**Steps**
1. Use the scene [SampleLogin - Connect Modal]
2. Login with remember me enabled
3. Stop the execution
4. Disable the remember me option
5. Try to login again

**Expected behavior**
User needs to restore the session again.

**Screenshots/Video**
https://github.com/user-attachments/assets/97361ead-3469-436d-b041-16ab01f02197

 - Chainsafe SDK version: [dev]
 - Unity Version: [2022.3.26]
 - Operating System: [Mac]

**Additional context**
Unless this is expected behaviour and the user needs to logout first.

rob1997:
closes #1103 
closes #1085


Packages/io.chainsafe.web3-unity/Runtime/Scripts/Connection/WalletConnectConnectionProvider.cs
```diff
@@ -11,36 +17,122 @@ namespace ChainSafe.Gaming.UnityPackage.Connection
     /// WalletConnect connection provider used for connecting to a wallet using WalletConnect.
     /// </summary>
     [CreateAssetMenu(menuName = "ChainSafe/Connection Provider/Wallet Connect", fileName = nameof(WalletConnectConnectionProvider))]
-    public class WalletConnectConnectionProvider : RestorableConnectionProvider
+    public class WalletConnectConnectionProvider : RestorableConnectionProvider, IWalletConnectConfig, IConnectionHandlerProvider
```
kantagara, line 20: Why is hyperplay connection provider not also the IConnectionHandlerProvider?

kantagara (approved):

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Connection/WalletConnectConnectionProvider.cs
rob1997: `IConnectionHandlerProvider` is specific to wallet connect, naming is a bit ambiguous

juans-chainsafe (changes requested):
Is working on the editor but I can't build on WebGL, getting the following error:
`/Users/juanmanuelspoleti/Desktop/workspace/web3.unity/Packages/io.chainsafe.web3-unity.hyperplay/Runtime/Scripts/HyperPlayWebGLProvider.cs(71,25): error CS1061: 'IHyperPlayConfig' does not contain a definition for 'RememberSession' and no accessible extension method 'RememberSession' accepting a first argument of type 'IHyperPlayConfig' could be found (are you missing a using directive or an assembly reference?)`

juans-chainsafe (approved):
Working now in WebGL as well!

creeppak (changes requested):
Looks good overall.

The only issue I have with this approach is that we force devs into using Connection Modal. Let's say somebody wants to use just one connection provider. Let's say it's WalletConnect. Now he won't be able to do it, without showing the connection modal with just one provider.
Maybe he even could, but now there is no configuration asset for WalletConnect (and likely for other providers too), as it was replaced with a ConnectionProvider. I think we should keep both approaches without forcing devs into using this one particularly.

We might make connection modal just instantly run the provider when there is only one available, but again, if somebody just wants to simply build a web3 without all this fancy stuff, he won't be able to do it and it's not good.

kantagara:
@creeppak good talking points for the tomorrows arch meeting. 
But, I think what Rob is doing is good. Because, 99% of the developers wouldn't mind doing things 'the way we intended them to work'

rob1997:
> Looks good overall.
> 
> The only issue I have with this approach is that we force devs into using Connection Modal. Let's say somebody wants to use just one connection provider. Let's say it's WalletConnect. Now he won't be able to do it, without showing the connection modal with just one provider. Maybe he even could, but now there is no configuration asset for WalletConnect (and likely for other providers too), as it was replaced with a ConnectionProvider. I think we should keep both approaches without forcing devs into using this one particularly.
> 
> We might make connection modal just instantly run the provider when there is only one available, but again, if somebody just wants to simply build a web3 without all this fancy stuff, he won't be able to do it and it's not good.

I've reverted the changes made and kept the config and `ConnectionHandlerProvider` in case some devs might want to do their own thing

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Connection/WalletConnectConnectionProvider.cs
kantagara: Yeah maybe change the name to `IWallectConnetConnectionHandler?`

creeppak (approved):
Good work.

PR #1118, opened 2024-08-29
Wallet connect on webgl

### LINKED ISSUE #1060: Wallet Connect WebGL Support (Especially Mobile)
**Summary**
An easy way for devs to connect their MetaMask wallet (or others) from mobile browser which doesn't have extension (ie. chrome)

I have tried using the web3modal package from wallet connect here
https://docs.walletconnect.com/appkit/unity/core/installation

I follow the installation document and just add a script (in pic) and a button with text in a scene (Unity 2022.3.10f1)
main

I build it in a new unity webgl project and test it using vscode live server from both my desktop and mobile device. It successfully connect the MetaMask wallet in both desktop and mobile browser, so I assume this implementation of Wallet Connect works in WebGL. It would be nice if this could be integrated into current ChainSafe SDK, so that Wallet Connect work on WebGL too, especially on mobile browser.

**Motivation**
While people can play WebGL game on desktop to connect with their wallet extension, I think it is crucial to support mobile devices too because it is more accessible to a lot of people. A lot of people spend times way more on mobile devices now, so this feature will be really helpful

**Acceptance Criteria**
Successfully connect a wallet from wallet app like MetaMask when opening WebGL build from mobile browser

**Additional Context**
Some screenshots of connecting wallet from mobile browser
image
image
image
image
image

**Final Steps**
* Give it one of these labels `Type: Feature` `Type: Enhancement`
* Add your estimation

**Estimations**
| 5 | Moderate efort | A few days | Mediium complexity | Moderate |

kantagara:
THere are some errors maybe on webGL, but everything seems to be working properly.
Closes #1060


juans-chainsafe (approved):
Samples are working with web3auth and Wallet connect! 

I found login and logout being veeery slow with wallet connect, at the point that I though that was not working, we need at least a loading spinner to know whats happening there @kantagara

kantagara:
@juans-chainsafe added the loading overlay, but only for when you are opening the wallet connect qr dialog (not when you actually connected, since I couldn't find a nice solution for that.) and when you are signing out.

juans-chainsafe (approved):
Spinner when WC modal is appearing and when you are logging out looks awesome!
True that we need one when is connecting to the wallet because of how much time it takes, but I believe this is great for now.
Great job.

juans-chainsafe (changes requested):
This happens to me because I saw the logging out spinner appear and approved the PR immediately, BUT the spinner doesn't disappear from the screen:
Screenshot 2024-09-03 at 1 23 44 PM

While the spinner is there, Im able to click buttons on the screen, even the logout button and if I click the logout button again, I'm redirected to the Login screen

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Web3Accessor.cs
```diff
@@ -46,8 +47,9 @@ public static async Task TerminateAndClear(bool logout = false)
             {
                 Debug.LogError("Instance was not set.");
             }
-
+            LoadingOverlay.ShowLoadingOverlay("Logging out...");
             await Instance.web3.TerminateAsync(logout);
+            LoadingOverlay.HideLoadingOverlay();
```
creeppak, line 52: I think we need to use finally statement each time we're using Show/HideLoadingOverlay(). This way it won't get stuck if TerminateAsync throws an exception.

creeppak: Or a custom using statement perhaps?

creeppak (changes requested):

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Web3Accessor.cs
creeppak: Ideally we should make loading overlay only show when used inside the custom `using` statement to enforce the 'safe' usage.

rob1997 (approved):

juans-chainsafe (approved):
Logout working as expected now!

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Web3Accessor.cs
creeppak: I mean that you can create a class, return an instance of this class in `ShowOverlay`, and then use the instance as an argument for `using (instance)` operator. This way when execution gets out of the using statement block (organically or cause of the exception) it will call `Dispose` on the instance.

The instance can be initialized with the reference of the overlay and will be able to call `HideOverlay` in `Dispose`. If you put the class of the instance in the Overlay class, it will even be able to call private methods of the overlay.

You can't really enforce using the `using` statement this way, but you can make it an obvious choice, idk

kantagara: Ah I see! 
But the loading overlay itself is a classic unitys singleton object, so I didn't want to enforce such use-cases for something as simple as loadingoverlay is.

creeppak: Well, as I see it. It's either **using with disposable** or you have to cover the 3 lines in **`try-finally`** each time you want to show a loading overlay. If not it will keep getting stuck.

kantagara: Yeah the latter seems better for now. We're not going to use it that extensively tho, only in two places basically.

PR #1135, opened 2024-09-12
Nikola/marketplace demo

kantagara:
@juans-chainsafe had to strip off a large portion of the marketplace sample scene (code is still there) just so that we can demonstrate how easy it is to list and purhcase the nfts. There is marketplace browser object where you can paste your project id, marketplace id and marketplace address, that will easily help you out to retreive items from the marketplace.


juans-chainsafe (changes requested):
merge conflicts

sneakzttv (approved):
Great additions, thank you for cleaning this up, conflicts RE Juans request above, otherwise good. We ran through this earlier today in the demo sprint, much nicer load speeds too.

juans-chainsafe (commented):
@kantagara also, be aware that MarketplaceSampleScene is set by default to open after the login in this PR, we should put back SampleMain

kantagara:
Closing this one in order to make the proper implementation that we already had in place.

PR #1150, opened 2024-09-17
Better way of adding dependencies'

rob1997:
This way we don't have to worry about what's already in `manifest.json`

### DIFF
--- a/Packages/io.chainsafe.web3-unity/Editor/WebGLThreadPatcherInstaller.cs
+++ b/Packages/io.chainsafe.web3-unity/Editor/WebGLThreadPatcherInstaller.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using UnityEditor;
 using UnityEngine;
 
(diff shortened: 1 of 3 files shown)


kantagara (approved):
Clever!

Packages/io.chainsafe.web3-unity/Editor/WebGLThreadPatcherInstaller.cs
```diff
@@ -36,7 +37,9 @@ static WebGLThreadPatcherInstaller()
     [MenuItem("ChainSafe SDK/Install WebGLThreadingPatcher", priority = 0)]
     public static void TryInstallThreadPatcher()
     {
-        Manifest manifest = JsonConvert.DeserializeObject<Manifest>(File.ReadAllText(ManifestPath));
```
KBryan, line 39: You would be better served wrapping this in a try/catch. Also, you should first check to ensure the ManifestPath exist before performing any operations.

rob1997: Manifest path is certain to exist since this is a Unity project, or else this script wouldn't run but I'll add the try catch...good catch xD

rob1997: changes made in latest commit

juans-chainsafe (approved):
didn't find build issues

KBryan (approved):
lgtm

sneakzttv (approved):

PR #1154, opened 2024-09-19
Single Scene

rob1997:


juans-chainsafe (changes requested):
- For some reason, in the editor, the Connect button doesn't appear 
- In WebGL, after you connect, some samples like GetNonce or GetTransactionStatus doesn't do anything. But others like Mint or Transfer works
- If I don't connect the wallet, and click a sample, I can't click anything else and I need to refresh the page, which is annoying. I should be able to see a error that wallet is not connected, and connect it.
- Web3Auth is not working: `openlogin-adapter:2 Failed to connect with openlogin provider WalletInitializationError: Invalid params passed in, PrivateKey Provider is required before initialization`
- I should be able to disconnect the wallet (I saw a disconnect button that appear and disappear one time, but looks like is not working properly)
- Spinner of Wallet Connect is not working, I refer to the spinner that Nikola added when the user is connecting the wallet
- What happens with MudSample scene or MarketplaceScene?

kantagara:
@juans-chainsafe MUD - eh, that boy uses local wallet, will be tricky
MarketplaceSample - actually I will wait for this PR to be finished so that I can make the proper implementation of it

rob1997:
> * If I don't connect the wallet, and click a sample, I can't click anything else and I need to refresh the page, which is annoying. I should be able to see a error that wallet is not connected, and connect it.

- I don't know why that's happening, please try finding `WalletConnectConnectionProvider` and clear cache then find `Application,PersistentDataPath` on your [device](https://docs.unity3d.com/ScriptReference/Application-persistentDataPath.html#:~:text=Mac%3A%20Application,on%20your%20system.) and delete everything in it. Let me know i that fixes it.
- Present on dev and also Editor but fixed
-  Fixed
- Present on dev
- Fixed
- Present on dev
- What was the flow like previously?

juans-chainsafe (changes requested):
Looking better!

- I was still not able to see the Connect button in the editor, so I can't test it there
- Disconnect button works now ✅ 
- I still able to click samples if the wallet is not connected (WebGL). I know that this happens in dev, but in dev you start in Connect modal scene, so you only go to SampleMain if you are connected.

juans-chainsafe:
> Looking better!
> 
> * I was still not able to see the Connect button in the editor, so I can't test it there
> * Disconnect button works now ✅
> * I still able to click samples if the wallet is not connected (WebGL). I know that this happens in dev, but in dev you start in Connect modal scene, so you only go to SampleMain if you are connected.

I see and it works, the Connect button inside a Desktop build

juans-chainsafe (approved):
- We decided to let the user click samples even if the wallet is not connected
- iOS working ✅ but new buttons are not displayed properly https://github.com/ChainSafe/web3.unity/issues/1159
- Editor working now ✅

sneakzttv (approved):
Thanks, this was tilting me

Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Token/Evm.cs
```diff
@@ -0,0 +1,178 @@
...
+using Nethereum.Hex.HexTypes;
+using Nethereum.Signer;
+using Nethereum.Util;
+using UnityEngine;
+using ChainSafe.Gaming.Marketplace;
+
+namespace Scripts.EVM.Token
+{
+    public static class Evm
+    {
```
kantagara, line 21: @rob1997 you can see which of these we have in Web3Unitys class. I think you should maybe just leave the GetNonce (or maybe, we can remove it from the samples all toghether). I did remove the evm class completely since I don't want people to reference it in 3.0 at all.

rob1997: There's no reference to `EVM` class in `Web3Unity.cs`

rob1997: Requested changes made

kantagara (approved):

PR #1158, opened 2024-09-20
Connect on Initialize Fix

rob1997:
There were issues in async execution order for web3auth in async plus some unhandled errors/exceptions

Bug was that connect would get stuck while trying to restore web3auth session


sneakzttv (approved):
fun way to pause for a frame, please add a description to the issue though., looks messy without it.

image

Packages/io.chainsafe.web3-unity.web3auth/Runtime/Web3AuthProvider.cs
```diff
@@ -93,7 +93,15 @@ private void Cancel()
...
     {
-        _connectTcs.SetResult(response);
+        if (string.IsNullOrEmpty(response.error))
+        {
+            _connectTcs.SetResult(response);
+        }
+
+        else
+        {
+            _connectTcs.SetException(new Web3Exception(response.error));
```
kantagara, line 103: Do we want to throw it in the users face? I am thinking we should silently fail but don't have strong opinion on it.🤷‍♂️

rob1997: when trying to restore session, it'll just log an error

kantagara (approved):

rob1997:
> Fun unity specific way to pause for a frame, please add a description to the issue though. It looks messy without it.
> 
> image

good call, I've updated the description of the PR

juans-chainsafe (approved):

rob1997:
Emm this was merged as part of #1154

Issue #1155, opened 2024-09-20
Web3Auth not working in WebGL


juans-chainsafe:
I'm not able to login with Web3Auth using a WebGL build. I get the following error:
`Failed to connect with openlogin provider WalletInitializationError: Invalid params passed in, PrivateKey Provider is required before initialization`

Is working in the editor.

rob1997:
fixed with #1170

PR #1161, opened 2024-09-25
Web3Auth bug fixes

rob1997:
- Moved out Web3Unity from ConnectToWallet because DontDestoryOnLoad don't work on parented objects
- Waiting a frame doesn't always work for fixing Web3Auth execution order so i adjusted via edits on their SDK
- We weren't unsubscribing to Login events a well, now we are


sneakzttv (approved):

juans-chainsafe (approved):
Looks good in the editor!
In WebGL Im getting this error https://github.com/ChainSafe/web3.unity/issues/1155

creeppak (approved):

PR #1163, opened 2024-09-25
Polling Event Manager - EVM events subscription for WebGL

creeppak:
## What's in the PR:
- We can now subscribe to EVM events on both multi-threaded and single-threaded platforms (WebGL). The interface is the same and can be accessed by devs using `web3.Events.Subscribe()`.
_For Web3 modules you can simply ask for a `IEventManager` dependency in the ctor just as before._
- Added a functionality that allows subscribing to events of the selected contract addresses, as it is an important part that was missing.
- Added `UseEventsWithWebSocket()`, `UseEventsWithPolling()`, and `UseEvents()` extension methods. The last one automatically determines which implementation it should register based on the current platform.
- Updated MUD module and samples a bit.

## To be tested:
Simply testing the MUD sample in build should be enough IMO, as MUD relies on EVM events heavily. We should test both WebGL and Standalone builds. The counter should get updated after pressing the 'Increment' button.


juans-chainsafe (changes requested):
I'm seeing the following build error:
`/Users/juanmanuelspoleti/Desktop/workspace/web3.unity/Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Events/EventExtensionsUnity.cs(3,33): error CS0234: The type or namespace name 'EventPoller' does not exist in the namespace 'ChainSafe.Gaming.Web3.Evm' (are you missing an assembly reference?)
`

creeppak:
> I'm seeing the following build error: `/Users/juanmanuelspoleti/Desktop/workspace/web3.unity/Packages/io.chainsafe.web3-unity/Runtime/Scripts/EVM/Events/EventExtensionsUnity.cs(3,33): error CS0234: The type or namespace name 'EventPoller' does not exist in the namespace 'ChainSafe.Gaming.Web3.Evm' (are you missing an assembly reference?) `

Messed up with refactoring a bit. Give me a sec..

creeppak:
@juans-chainsafe Should be alright now.

juans-chainsafe (changes requested):
I found 1 thing to fix:

- In a WebGL build, the number is not incremented, even if you increase it on WebGL but is increased in the mud localhost

juans-chainsafe (changes requested):
Still the counter is not incremented on WebGL:

Screenshot 2024-09-30 at 9 55 59 AM

Screenshot 2024-09-30 at 9 56 27 AM

This is the log when I click **Increment Counter** button `Sending transaction to execute the Increment function.. `

Is updated only if I refresh the page

kantagara (approved):

creeppak:
@juans-chainsafe  The problem was related to WebGLThreadingPatcher. As we added it back, it all started working as expected in builds.

PR #1167, opened 2024-09-30
Read-only Web3

rob1997:
run `setup.sh`


juans-chainsafe (approved):
Looks good! the only downgrade is if the tx gets stuck, the loading spinner can't be closed, so you need to restart the scene

rob1997:
> @rob1997 can you maybe add the close button on the loading screen overlay there then?

That sort of defeats the whole purpose of a loading overlay, doesn't it?

kantagara:
@rob1997 True 😁  but maybe after certain amount of seconds it can appear so that users are not completely blocked? Or we can just timeout the loading spinner itself to like 5-10 secs then say this transaction takes longer than usual, it will be processed in the background, you can close this loading window (and add the X then)

rob1997:
> @rob1997 True 😁 but maybe after certain amount of seconds it can appear so that users are not completely blocked? Or we can just timeout the loading spinner itself to like 5-10 secs then say this transaction takes longer than usual, it will be processed in the background, you can close this loading window (and add the X then)

I can put a timeout that'll cancel the task which'll close the overlay

kantagara (approved):

sneakzttv:
awaiting conflict resolution

sneakzttv (approved):

PR #1177, opened 2024-10-03
3.0.0 everywhere

kantagara:


creeppak:
Don't we need to update version in each `package.json` of each package?

creeppak:
Or wait, they were updated? Probably missed while viewing the changes.

PR #1182, opened 2024-10-07
Marketplace Sample scene

kantagara:
Marketplace sample now actually just consist of the ability to browse a designated marketplace collection, whilst still utilizing our beloved Service Adapters. 

### I apologize for so many files changed, linter is to blame!!


Packages/io.chainsafe.web3-unity.web3auth/Runtime/chainsafe.web3-unity.web3auth.asmdef.meta
```diff
@@ -1,11 +1,6 @@
 fileFormatVersion: 2
-<<<<<<<< HEAD:src/UnitySampleProject/Assets/Scenes/SampleMain/Category.prefab.meta
 guid: 065b2a8044c813947b0e4d52126f5a26
-PrefabImporter:
-========
-guid: d0e5603e6af957942922f5d22bbbfa0c
```
rob1997, line 6: what is this removed assembly?

kantagara: unity must've done it. lemme see

kantagara: image
everything seems to be alrightt there

Packages/io.chainsafe.web3-unity.marketplace/Editor/AddMarketplaceScriptingDefineSymbol.cs
```diff
@@ -0,0 +1,16 @@
...
+
+[InitializeOnLoad]
+public class AddMarketplaceScriptingDefineSymbol
+{
+    private const string DefineSymbol = "MARKETPLACE_AVAILABLE";
+    private static bool initialized;
+
+    static AddMarketplaceScriptingDefineSymbol()
+    {
+        ScriptingDefineSymbols.TryAddDefineSymbol(DefineSymbol);
```
rob1997, line 13: Do we still need this? Do we have any marketplace scripts in core pkg anymore? if so can't we use a combination of `ISample` and service adapters? same as ramp

rob1997 (approved):
linter changes and scene changes will probably cause conflicts

sneakzttv (approved):

Packages/io.chainsafe.web3-unity.marketplace/Editor/AddMarketplaceScriptingDefineSymbol.cs
sneakzttv: dont think so, it was for conditional for the sample scene display.

src/ChainSafe.Gaming.Marketplace/MarketplaceClient.cs
```diff
@@ -70,15 +57,13 @@ public MarketplaceClient(IHttpClient httpClient, IProjectConfig projectConfig, I
         /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         public Task<MarketplacePage> LoadPage(
             MarketplacePage? currentPage = null,
-            string? marketplaceId = null,
```
creeppak, line 73: Do we need to force user into using just one marketplace in his game?

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Connection/ServiceAdapters/MarketplaceServiceAdapter.cs
```diff
@@ -7,14 +10,33 @@ namespace ChainSafe.Gaming.UnityPackage.Connection
...
+                services.UseMarketplace(marketplaceConfig);
             });
         }
     }
+
+    [Serializable]
+    public class MarketplaceConfigUnity : IMarketplaceConfig
+    {
+        [field: SerializeField] public string EndpointOverride { get; set; }
+        [field: SerializeField] public string MarketplaceContractAbi { get; set; }
```
creeppak, line 35: This can trick people into thinking they need to provide ABI for their marketplace while it's actually being loaded by default. One possible way to fix is adding "Override" prefix to the unity field name.

src/ChainSafe.Gaming.Marketplace/MarketplaceConfig.cs
```diff
@@ -2,34 +2,26 @@
...
-            EndpointOverride = "https://api.gaming.chainsafe.io",
-            MarketplaceContractAbi = ReadDefaultAbiFromResources(),
-        };
-
         /// <inheritdoc/>
         public string? EndpointOverride { get; set; }
 
         /// <inheritdoc/>
-        public string? MarketplaceContractAbi { get; set; }
+        public string MarketplaceContractAbi { get; set; }
```
creeppak, line 19: Could rename this into `MarketplaceContractAbiOverride` and add another readonly property that will either use the override when it's not null or `ReadDefaultAbiFromResources`. All of this can sit happily in `IMarketplaceConfig`.

creeppak (changes requested):

creeppak:
Marketplace samples yey!

src/ChainSafe.Gaming.Marketplace/MarketplaceClient.cs
kantagara: For now I'd say so.
I mean _techcnically_ , we should be able to easily add the ability for users to load multiple marketplaces, but let's do it post 3.0. I will remove the marketplaceId from there since it doesn't make sense to be in the LoadPage

Packages/io.chainsafe.web3-unity.marketplace/Editor/AddMarketplaceScriptingDefineSymbol.cs
rob1997: yeah I got that we can still use a combo of the two like we did for Ramp samples

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Connection/ServiceAdapters/MarketplaceServiceAdapter.cs
kantagara: Fixed

Packages/io.chainsafe.web3-unity.marketplace/Editor/AddMarketplaceScriptingDefineSymbol.cs
kantagara: Marketplace sample scene is a standalone scene of it's own now that we have the ability to have a single scene system. Same should be done for MUD imo.

PR #1200, opened 2024-10-21
Adding the event to the web3unity so that anyone anywhere can be informed when web3 instance inside of the Web3Unity is initialized.

kantagara:


creeppak (approved):

rob1997 (approved):

kantagara:
@juans-chainsafe not much to see here tbh since this doesn't have any direct influence on the code execution.

juans-chainsafe (approved):

PR #1212, opened 2024-10-31
Added bool for event polling to not enfroce people to use websockets.

kantagara:
Updated rider plugin.
Made Multicall not throw exception if the network doesn't support multicalls.


rob1997 (approved):

Packages/io.chainsafe.web3-unity/Runtime/Scripts/Connection/ServiceAdapters/EventsServiceAdapter.cs
```diff
@@ -8,14 +8,22 @@ namespace ChainSafe.Gaming.UnityPackage.Connection
 {
     public class EventsServiceAdapter : MonoBehaviour, ILightWeightServiceAdapter
     {
+        [SerializeField, Tooltip("Enable this if a WebSocket connection is unavailable or if the blockchain does not provide a WebSocket URL")]
+        private bool forceEventPolling;
```
creeppak, line 12: Question: Why do we make these serialized fields private? I don't think it makes it more safe to use, but it does make it impossible for devs to set this value programmatically.

Don't get me wrong there are cases where we'd want to make the value only editable through the editor ui, but I don't think this is one of those cases.

kantagara: But this is only the 'facade' that is used to inject the event service adapter.

The ForceEVentPolling is public in the config itself, so devs can _always_ inject the event service manually and configure it however they want. They shouldn't seek for the EventsServiceAdapter to do that type of configuration at all, if they're doing everything 'by hand'.

creeppak (approved):

juans-chainsafe (approved):
- Error log appears if multicall is not supported
`Web3Exception: Can't use MultiCall. No contract for the active chain is available.`

- I can login without websocket with the Force Event Polling checked, otherwise throws error as expected.
Screenshot 2024-10-31 at 9 44 37 AM

PR #1225, opened 2024-11-05
We now dispose services more properly

### LINKED ISSUE #1211: Stuck in Reloading Domain when Web3Auth login is cancelled and EventsServiceAdapter is present
**Describe the bug**
If you add EventsServiceAdapter script into Web3Unity object and try to login with web3auth but you close the login modal, stop the execution and hit play again in the editor, then you get stuck in Reloading Domain loading popup.

**Steps**
1. Add **EventsServiceAdapter** on Web3Unity GameObject 
Screenshot 2024-10-29 at 5 19 18 PM

2. Start SampleMain in the editor (disconnect wallet if is connected)
3. Click on **Connect** button
4. Select Web3Auth option
5. Select Google option
6. Close the Web3Auth browser tab (don't connect)
7. Go back to unity and close the connection modal
8. Stop the execution (in the editor)
9. Start the execution (click play button in the editor)

**Expected behavior**
SampleMain loads correctly and user can try to login again

**Screenshots/Video**
Screenshot 2024-10-29 at 12 38 16 PM

 - Chainsafe SDK version: [3.0.4]
 - Unity Version: [2022.3.44]
 - Operating System: [Mac]

**Additional context**
Workaround is to close Unity editor and open it again.

rob1997:
All started lifecycle participants will be stopped even if web3 fails to initialize
Use `[ExecutionOrder(int)]` attribute on services implementing `ILifecycleParticipant`
`WillStopAsync` also calls for all `ILifecycleParticipant`s even if an exception is thrown.
closes #1211


src/ChainSafe.Gaming/RPC/Events/WebSocketEventManager.cs
```diff
@@ -14,6 +14,10 @@
 
 namespace ChainSafe.Gaming.RPC.Events
 {
+    /// <summary>
+    /// <see cref="ILifecycleParticipant.WillStartAsync"/> will be executed later.
+    /// </summary>
+    [ExecutionOrder(10)]
```
kantagara, line 20: why is this an attribute and not a member that you then just use for sorting services?

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// </summary>
+    public class LifecycleHandler
+    {
+        private readonly ILifecycleParticipant[] lifecycleParticipants;
+
+        public LifecycleHandler(IEnumerable<ILifecycleParticipant> lifecycleParticipants)
+        {
+            // Arrange execution based on ExecutionOrder Attribute priority.
+            this.lifecycleParticipants = lifecycleParticipants
+                .OrderBy(p => p.GetType().GetCustomAttribute<ExecutionOrderAttribute>()?.Order ?? 0)
```
kantagara, line 20: Yeah instead of a reflection why we didn't use the field? Since we do support default interface implementation in the core solution.

src/ChainSafe.Gaming/RPC/Events/WebSocketEventManager.cs
rob1997: it's essentially the same thing but this is less invasive, we just have to put this attribute and if it's not present it'll default the execution order to 0.

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
rob1997: Interfaces can't hold data by default just implementations/logic, so every concrete implementation will have to implement the `ExecutionOrder` property

kantagara: image
But you can do this.

kantagara: And if you need to override it you can do that in the implementator later on 
image

rob1997: I had no idea, C# reversing entropy out here xD

rob1997: I'll redo it

src/ChainSafe.Gaming/RPC/Events/WebSocketEventManager.cs
rob1997: redone

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
rob1997: done

juans-chainsafe (approved):
Working! Now you can not connect to web3auth, stop the execution and start it again.
Tested with Force Event Polling on and off

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (exceptions.Count != 0)
+            {
+                throw new AggregateException(exceptions);
+            }
+
+            async Task StopAllParticipantsAsync()
+            {
+                try
+                {
+                    foreach (var participant in lifecycleParticipants.Where(p => !stoppedParticipants.Contains(p)))
```
creeppak, line 73: You could simply create a queue, initialize it with all the `lifecycleParticipants` and `while (queue.Count > 0)` them instead of using recursion. You could then also remove the `.Where(p => !stoppedParticipants.Contains(p)`.

Another important thing is that you should terminate things in the opposite direction comparing to how you initialize them. This means that if some service had to be initialized last (due to its execution order), it should terminate first. This is for example important if this service depends on initialization of another one. So if service B need to initialize after service A (A, B), it should terminate before service A (B, A), as B is dependent on A and A is not dependent on B.

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace ChainSafe.Gaming.Web3.Core
+{
+    /// <summary>
+    /// Handles the lifecycle of <see cref="ILifecycleParticipant"/> instances.
+    /// </summary>
+    public class LifecycleHandler
```
creeppak, line 12: Looks more like `LifecycleManager` to me as it manages all the `ILifecycleParticipant`s.

creeppak (changes requested):
All good, just needs a couple of improvements.

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
sneakzttv: decent advice....removing dependency here is good.

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
sneakzttv: handler/manager seem interchangeable to me but if we have it like this elsewhere and we're sticking with uniformity then yeah, change it.

sneakzttv (approved):

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
rob1997: handler, manager...potato, powtatow xD will change

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
rob1997: ahh good point, will do

rob1997: done

src/ChainSafe.Gaming/Web3/Core/LifecycleHandler.cs
rob1997: done

PR #1246, opened 2024-11-27
Sync Branches CI fix

rob1997:
- Fixes merging main to dev automatically
- What it does is basically whenever we push to main it tries and merge it to dev (sync both branches)
- we're using Setup.sln
~- Note :- It doesn't resolve conflicts and we're likely to run into them since we're updating dependencies via CI on both branches so at the very least library folders with dlls will have conflict~ Now we got an exclusion path list
- Improved command line parsing for adding new option in Setup.sln
- Updated label checks in PR to dev to work on simultaneous labeling
- as usual CI won't work till this is on main but [here's](https://github.com/ChainSafe/web3.unity/actions/runs/12069815356/job/33657900672) a successful test run

### DIFF
--- a/.github/workflows/duplicate_samples.yaml
+++ b/.github/workflows/duplicate_samples.yaml
@@ -3,25 +3,12 @@ name: Duplicate Samples 🪞
 on:
   workflow_call:
   workflow_dispatch:
-    inputs:
-      arguments:
-        required: false
-        description: 'Run Arguments'
-        type: string
-      configuration:
-        required: true
-        description: 'Run Configuration'
-        default: 'Release'
-        type: choice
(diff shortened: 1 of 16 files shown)


kantagara:
Why can't we force the things from dev to be the defaultly accepted if there is a conflict? At least for dlls

rob1997:
> Why can't we force the things from dev to be the defaultly accepted if there is a conflict? At least for dlls

Yeah we can, I'll need to add it to the `Setup.sln` though for access to lib paths...adds some effort

kantagara (approved):

sneakzttv (approved):

PR #1256, opened 2024-12-11
Toast Overlay Type Implemented

### LINKED ISSUE #1178: Add Toast to Connect if you need the write call.
**Summary**
Add toast message to the Connect modal to show why it popped up if user decides to execute a write call whilst he didn't connect his wallet

rob1997:
closes #1178


Packages/io.chainsafe.web3-unity/Runtime/GUI/GuiInfoOverlay.cs
```diff
@@ -8,14 +8,18 @@ namespace ChainSafe.Gaming.GUI
     public class GuiInfoOverlay : MonoBehaviour
     {
         public TMP_Text Message;
+        public TMP_Text ToastMessage;
```
creeppak, line 11: Why not reuse 'Message'?

rob1997: because toast message has a different positioning (bottom) and is under a layout group that has a `MaxSize` component on it

juans-chainsafe (changes requested):
I don't see the modal in all the samples, only in TransferERC20.
For example in Mint20or Mint721 or TransferERC721, the modal should appear

rob1997:
> I don't see the modal in all the samples, only in TransferERC20. For example in Mint20or Mint721 or TransferERC721, the modal should appear

to be fair this is also present on dev but I'll fix it

rob1997:
> > I don't see the modal in all the samples, only in TransferERC20. For example in Mint20or Mint721 or TransferERC721, the modal should appear
> 
> to be fair this is also present on dev but I'll fix it

fixed, make sure you run `setup.sh`

juans-chainsafe (approved):
now is working for all samples that require a provider! Also working in WebGL

sneakzttv (approved):

kantagara (approved):

creeppak (approved):

Issue #1262, opened 2024-12-13
Error with Toggle chain sample when you switch back to the original chain with Metamask on WebGL


juans-chainsafe:
**Describe the bug**
After you toggle the chain for second time, you get a new error using Metamask in WebGL

**Steps**
1. Build on WebGL
2. Connect to Metamask
3. Click on Toggle chain button
4. Click again on Toggle chain button

**Expected behavior**
Chain is switched back to primary user chain set on project settings

**Console errors**
```
inpage.js:1 MetaMask - RPC Error: nativeCurrency.symbol does not match currency symbol for a network the user already has added with the same chainId. Received:
ETH 
Object
Build.framework.js:3013 [Web3] Failed to add or switch to the network: nativeCurrency.symbol does not match currency symbol for a network the user already has added with the same chainId. Received:
ETH
Build.framework.js:3013 Web3Exception: nativeCurrency.symbol does not match currency symbol for a network the user already has added with the same chainId. Received:
ETH
Rethrow as InvalidOperationException: Failed to add or switch to the network.
  at ChainSafe.Gaming.Web3.Evm.Wallet.WalletProvider.SwitchChainAddIfMissing (ChainSafe.Gaming.Web3.IChainConfig config) [0x00000] in <00000000000000000000000000000000>:0 
Rethrow as Web3Exception: Error occured while trying to switch wallet chain.
  at ChainSafe.Gaming.Web3.Evm.Wallet.WalletProvider.HandleChainSwitching () [0x00000] in <00000000000000000000000000000000>:0 
Rethrow as Web3Exception: One of the handlers thrown an exception. Reverted ChainManager to the previous chain configuration.
  at ChainSafe.Gaming.Web3.Core.Chains.ChainManager.SwitchChain (System.String newChainId) [0x00000] in <00000000000000000000000000000000>:0 
```

**Desktop (please complete the following information):**
 - Chainsafe SDK version: [dev]
 - Unity Version: [2022.3.44]
 - Operating System: [Mac]
 - Platfowm: [WebGL]

PR #1263, opened 2024-12-16
Error with Toggle chain sample when you switch back to the original chain with Metamask on WebGL #1262

creeppak:
- Moved "SwitchChainAddIfMissing" logic to be used in Reown only.
- Switching chains the old simple way for all the Wallet Providers by default.

A new bug related to Reown was discovered. You can't switch chains in Reown as exception is getting thrown. It also happens in dev. 

The error: 

```
MissingMethodException: Constructor on type 'ChainSafe.Gaming.Reown.Methods.WalletAddEthereumChain' not found.
```


rob1997 (approved):
Nice! however this could conflict with #1270

kantagara:
Closed sine metamask will be removed in 3.1 in favour of appkit

PR #1277, opened 2025-01-22
CI Fix on dev

rob1997:
We were initializing web3 twice on dev which was causing all sorts of issues because `OnWeb3Initialized` callback was calling twice, here's a test action [run](https://github.com/ChainSafe/web3.unity/actions/runs/12906995469/job/35989517093)


kantagara (approved):
One question tho: Wouldn't this cause some issues in runtime? Since lightweight instance/heavyweight instance wouldn't be initialized if rememberConnection is set to false? Oir they wil be?

rob1997:
> One question tho: Wouldn't this cause some issues in runtime? Since lightweight instance/heavyweight instance wouldn't be initialized if rememberConnection is set to false? Oir they wil be?

`rememberConnection` is used to restore Web3 if a connection exists and if not we'll still initialize lightweight web3

juans-chainsafe (approved):

